PR #2, opened 2018-08-22
Adding a task to autofix

nataliecarey:
Autofix exists for `standard` out of the box and there's a module for autofixing `sass-lint`.  This makes both of those available under the command:

```
npm run test:autofix
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1140,6 +1140,12 @@
       "integrity": "sha1-tUc7M9yXxCTl2Y3IfVXU2KKci/I=",
       "dev": true
     },
+    "charenc": {
+      "version": "0.0.2",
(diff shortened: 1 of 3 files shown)


rpowis:
i'm not sold on adding this as an npm script. it's adding a dependency that'll only really be needed when porting lots of code from a project without `sass-lint` or different `sass-lint` rules. this can be run ad-hoc in those instances with `npx sass-lint-auto-fix` and we can add instructions to our (currently non-existent 😟) README to do that.

same with `npx standard --fix` if going through all the issues manually takes too long. if people really want to be running `--fix` every time then they can set up their editors to do so. again, something for the README.

code written during dev should be being linted as it's written, if it isn't then that's something we can look at doing.

PR #3, opened 2018-10-10
Account menu tests

rpowis:
This PR covers:

* Removing jQuery as a dependency of account-menu 🎉 
* Adding tests for the account-menu component. Tests are run with jest, they use puppeteer for browser interaction, and jest-axe for testing accessibility (as per govuk-frontend).
* We're now building and serving an app to point puppeteer at for the tests. This could do with a bit of work (no global JS initialisation file, no offline capability, etc.), but as a future iteration not part of this PR.
* Watch and serve tasks for faster development
* A few config files (eslint and editorconfig)
* Updated test dependencies

## Notes

This one's a biggie. Sorry about that. We could probably split it into a few PRs if so desired (e.g. one for the config and dependency stuff, one for the puppeteer app, and one for the account-menu work). There are also a few commits that can be squashed in the `account-menu.js` file, but I figured that would make the changes harder to see.

I went down a few worm holes with the JS for the account menu so ended up reduced the scope to removing jQuery and nothing more. We need to discuss how closely we align with `govuk-frontend`'s JS patterns, but either way there's more work to be done on the account menu which we can raise issues for (styles, markup, and JS).

Similarly, the app is a first pass too. What's there is enough to get the tests working and was taken as-is from `govuk-frontend` so there are a few extra that I don't think are necessary, but again we can raise issues around that too.


rpowis:
Ah yes, that merge conflict has reminded me that the images are all returning `404`s for the account menu. That's another improvement to be made to the app.

PR #11, opened 2018-11-22
SDT-879 Add Account header

rpowis:
Adding the Account Header as a layout file and styles. 

image

There needs to be some follow up work to this in order to make the language select more dynamic and to add some tests for the layout file, but that would benefit from further discussion first.


src/components/account-menu/template.njk
```diff
@@ -9,7 +11,10 @@
       <a href="#" class="hmrc-account-menu__link">Back</a>
     </li>
     <li>
-      <a href="#" class="hmrc-account-menu__link">Messages<span class="hmrc-notification-badge">2</span>
+      <a href="#" class="hmrc-account-menu__link">Messages
```
nataliecarey, line 14: By introducing a line break you're introducing a space into the content - I think that's a good thing, just want to make sure you're aware of this change.

src/components/notification-badge/notification-badge.yaml
```diff
@@ -1,7 +1,7 @@
 examples:
 - name: default
   data:
-    text: 3
+    badgeText: 3
 - name: with-text
   data:
-    text: New
+    badgeText: New
```
nataliecarey: Is this related to the account header? What's the reason for breaking the `text` convention govuk are using?

src/components/account-menu/template.njk
rpowis: yep, that was the idea 😄 it's a much tidier way of doing it than having a space followed by the conditional all on one line

app/app.js
```diff
@@ -12,7 +12,8 @@ const appViews = [
   configPaths.layouts,
   configPaths.views,
   configPaths.components,
-  configPaths.src
+  configPaths.src,
+  configPaths.govukFrontend + '/components'
```
nataliecarey, line 16: What problem does this solve?

PR #13, opened 2018-12-05
Page heading

rpowis:
## Default

image

## With Section

image


tasks/gulp/compile-assets.js
```diff
@@ -34,10 +34,14 @@ const errorHandler = function (error) {
 const compileStyleshet = isDist ? configPaths.src + 'all.scss' : configPaths.app + 'assets/scss/app.scss'
 const compileOldIeStyleshet = isDist ? configPaths.src + 'all-ie8.scss' : configPaths.app + 'assets/scss/app-ie8.scss'
 
+const sassConfig = {
+  includePaths: ['node_modules']
```
nataliecarey, line 38: Scary.  Are we sure we want to do this?

gordonmcmullan: That was my initial thought too, but I went and read the documentation I could find for `includePaths`. I hope I interpreted it correctly, this just sets `node_modules` as an additional search path to `.` for sass `@import` (and similar?) statements. so that things like `@import "govuk-frontend/tools/px-to-em";` will fetch the correct things.

PR #18, opened 2019-01-31
Sdt 978 review header

nataliecarey:


nataliecarey:
I'm happy with these changes, I'm concerned about the `#` default on the links but this change enhances the previous version so I think we can look later.

PR #19, opened 2019-02-07
SDT-1005 Fixed styling of the language selector links in the Account Header

gordonmcmullan:


rpowis:
this should be opened against #18 instead 👍

PR #23, opened 2019-02-14
Removing legacy safety-check

nataliecarey:
This safety-check was introduced to keep backwards compatibility in the examples, now that the extensions are part of gouvk-prototype-kit and this variable is part of the contract/interface it is unnecessary. 

I'd like to be able to point people to this file as an example for them to follow, I think that example is best without the safety-check.

### DIFF
--- a/src/_prototype-kit.scss
+++ b/src/_prototype-kit.scss
@@ -1,5 +1,3 @@
-$govuk-extensions-url-context: "/extension-assets" !default;
-
 $hmrc-assets-path: $govuk-extensions-url-context + "/hmrc-frontend/";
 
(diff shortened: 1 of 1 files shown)


rpowis:
is it not needed if hmrc-frontend gets installed into a pre-extensions version of the prototype kit?

nataliecarey:
Oh, maybe.  Darn.

rpowis:
but maybe we can leave it in but commented out with a comment above if it's needed pre-extensions?

rpowis:
i'm with you though. maybe just stick a comment above it explaining that it's only needed in that circumstance

PR #24, opened 2019-02-14
Update account menu to allow defining active link

rpowis:
This allows the account menu items to be set as active.

* Only account menu, messages, and check progress can be set as active
* The account menu defaults to active if none are explicitly set


nataliecarey:
OK, on our side this looks good, what's the likely impact on users?  Dynamically/conditionally setting a field on an object inside an array is more complex.  As these items all have names should we use something more like `active: "accountHome"`?

To be clear - I think this interface is great for our side of the contract but I'm wondering how easy people will find it to consume.

PR #34, opened 2019-03-15
SDT-884 Add to a list component

gordonmcmullan:
## Create a component for the Summary List used for items in the Add to a List pattern.
Realise the unique part of the Add to a List pattern as a component for use in prototypes
### Decisions: 
- Include the form used to initiate the journey to add another item.
- Depart from the api used in the GOVUK Frontend summary list component, they require more flexibility than we do in this particular pattern
- Like the other patterns we have already completed this doesn't include Welsh Language support 

### Default appearance with an Empty List
Screenshot from 2019-03-15 10-40-19

### With an single item using the default item name
Screenshot from 2019-03-15 10-41-53

### With multiple items named director/directors and some hint text
Screenshot from 2019-03-15 10-41-32

### DIFF
--- a/lib/jest-helpers.js
+++ b/lib/jest-helpers.js
@@ -9,10 +9,9 @@ const yaml = require('js-yaml')
 const configPaths = require('../config/paths.json')
 const { componentNameToMacroName } = require('./helper-functions.js')
 
-nunjucks.configure(configPaths.components, {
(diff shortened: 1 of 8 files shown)


gordonmcmullan:
'Do you need to add another item?' Might want reworded for an empty list

PR #37, opened 2019-04-04
Sdt 882 timeout dialog

nataliecarey:


src/components/_all.scss
```diff
@@ -4,4 +4,4 @@
 @import "hmrc-language-select/language-select";
 @import "hmrc-header/header";
 @import "hmrc-internal-header/internal-header";
-@import "hmrc-add-to-a-list/add-to-a-list";
```
gordonmcmullan: I don't think this should be removed

PR #60, opened 2020-02-06
PLATUI-342: Distribute compiled CSS inside the package.

nataliecarey:
Now that `hmrc-frontend` is used directly in Scala frontends the SCSS compilation was becoming difficult to support.  It makes sense to provide compiled files rather than files which require compilation.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -129,6 +129,7 @@ gulp.task('build:package', cb => {
     'copy-govuk-config',
     'copy-check-compatibility',
     'js:compile',
+    'copy-assets',
     'copy:README',
     'copy:packageJson',
     cb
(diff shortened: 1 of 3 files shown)


vinnicombe:
The problems I foresee with this approach is that it makes all services that consume these assets reliant on a single CDN resource. Multiple services with a single point of failure for assets is not good and has been a problem in the past. When the CDN that served Template as a Service, used by all personal tax frontends, became unavailable and the caching of the assets expired, it took 18 services down for hours.

vinnicombe:
Further to this, serving compiled assets limits the way services can tailor the assets they pull down. It removes the modular aspect of the library.

vinnicombe:
It is not complicated to make this work for Scala frontends by using relative paths for sass imports, I have tested a solution in a prototype and Scala FE. The solution is on a branch in my fork, I can make this available. Maybe providing both of these solutions would be the best way forward? Relative import paths will not change the compiled assets and will make the dependency published on mvn repository work in a Scala Play application.

PR #62, opened 2020-03-02
Added a basic component for reporting a technical issue.

nataliecarey:
As teams move over to using `govuk-frontend` and `hmrc-frontend` in production there's a need to add the technical issue form to each page of each service.

The current implementation has a few issues with it which are currently being reviewed.  I see the following options:

1. Don't create a component until we have a new recommendation of how this should be implemented
2. Create a new component with our best attempt at improving the experience
3. Create an exact copy of what's currently live, making it compatible with `hmrc-frontend`
4. Create a component which uses the non-js behaviour of the current live implementation for teams to use

This PR is using option 4 which I believe is best.  I ruled out option 1 because it will lead to lots of duplicate code which will be difficult to replace, it will hinder the consistency of the site.  I ruled out option 2 because I don't believe there is consensus on what the "best" option is right now.  I ruled out option 3 because there are lots of issues which would take time to replicate and we know it requires improvement.

I went with option 4 because it allows teams to use the component in its simplest form and then upgrade `hmrc-frontend` to get the latest version when we make improvements.  It gets teams moving quickly now *and* makes a quicker path to production for whatever the working group settle on as the best next version.

This supports english and welsh, it has tests.  It's simple but I believe it's worthy of being a component based on how widely it's reused.


src/components/report-technical-issue/template.njk
```diff
@@ -0,0 +1,2 @@
+<a class="govuk-link {{ params.classList }}" target="_blank" href="/contact/problem_reports_nonjs?service={{ params.serviceCode | default("unknown") }}"
+>{% if params.language == 'cy' %}Help gyda'r dudalen hon{% else %}Get help with this page{% endif %}</a>
```
timsb: Is there any reason you've chosen this wording as opposed to some of the options in services such as 'Is there anything wrong with this page?' or 'Is this page not working properly?'.?

src/components/report-technical-issue/template.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+    it('uses the provided URL', () => {
+      const $ = render('report-technical-issue', examples.welsh)
+
+      const $component = $('.govuk-link')
+
+      expect($component.text()).toEqual('Help gyda\'r dudalen hon')
+    })
+  })
+})
```
timsb, line 67: There should be a test that tests the `?service=unknown` query string.

src/components/report-technical-issue/template.njk
```diff
@@ -0,0 +1,2 @@
+<a class="govuk-link {{ params.classList }}" target="_blank" href="/contact/problem_reports_nonjs?service={{ params.serviceCode | default("unknown") }}"
```
timsb: - Have you thought about the change in behaviour with the form opening in a new tab?
- Generally we add some text in parenthesis to help users know that this is going to open in a new window or tab

timsb: Is `rel="noopener noreferrer"`needed as we're opening in a new tab? I think the answer is no but thought I'd raise anyway.

timsb (changes requested):

src/components/report-technical-issue/template.njk
nataliecarey: The reason is to encourage this conversation. If there's an "approved" new set of wording I'd love to include it.

src/components/report-technical-issue/template.test.js
```diff
@@ -0,0 +1,45 @@
...
+      expect(results).toHaveNoViolations()
+    })
+
+    it('renders a link element', () => {
+      const $ = render('report-technical-issue', examples.default)
+
+      const $component = $('.govuk-link')
+      expect($component.get(0).tagName).toEqual('a')
+      expect($component.attr('target')).toEqual('_blank')
+      expect($component.attr('rel')).toEqual(undefined)
```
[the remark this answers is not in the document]
nataliecarey: I was looking at this in relation to the `new-tab-link`, that specifically runs with `noopener noreferrer` which I feel are actively inappropriate for this usage.  I wanted to document that it shouldn't have those.  If we agree that it shouldn't have a `rel` it might be worth putting that in its own test and naming it more clearly.

src/components/report-technical-issue/report-technical-issue.yaml
```diff
@@ -0,0 +1,15 @@
+examples:
```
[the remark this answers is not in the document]
nataliecarey: Yes they should, I'll add them.

src/components/report-technical-issue/template.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+    it('renders link with custom classes', () => {
+      const $ = render('report-technical-issue', examples['with-classes'])
+
+      const $component = $('.govuk-link')
+
+      expect($component.attr('class')).toEqual('govuk-link govuk-!-font-weight-bold my-custom-class')
+    })
+
+    it('uses the provided URL', () => {
```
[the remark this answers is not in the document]
nataliecarey: Yes it should :( I'll update it.

src/components/report-technical-issue/README.md
```diff
@@ -0,0 +1,7 @@
+# Open a link in a new window or tab
```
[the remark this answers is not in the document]
nataliecarey: Doh, I'll update.

src/components/report-technical-issue/template.njk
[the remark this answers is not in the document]
nataliecarey: I'm happy with that.  I also considered not rendering the link at all if the service is not provided.  If we could fire errors I'd error if a service isn't provided, in other examples (like `new-tab-link`) we've displayed nothing if it's missing a required field.

src/components/report-technical-issue/template.test.js
nataliecarey: Yes, whatever behaviour we're using for that case should be documented in a test.

src/components/report-technical-issue/template.njk
nataliecarey: That's fair.  I'll update that.

nataliecarey: It's important if you're linking to external pages, this one is still inside HMRC therefore the security issue doesn't really exist.  I think the non-js version picks up information from the referrer and adds it to the information logged with the ticket, if they may do that then I'd see a benefit in passing that information through.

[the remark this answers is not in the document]
nataliecarey: Fair

src/components/report-technical-issue/template.njk
nataliecarey: The default should be changed to a blank string.

nataliecarey: Add query parameter `newTab=true` - this will be ignored but will allow us to update the messaging at a later date.

timsb (dismissed):

timsb (dismissed):

check-compatibility.js
```diff
@@ -12,6 +12,9 @@ if (!knownPrototypeKitNames.includes(consumerPackageJson.name)) {
 }
 
 const compatibility = {
+  '1.10': {
```
timsb, line 15: I think you just need to replace 1.9 rather than adding in a new one

timsb (approved):

PR #63, opened 2020-03-30
Relative paths

vinnicombe:


vinnicombe:
@natcarey @timsb @adamliptrot-oc please see PR for readme updates

README.md
```diff
@@ -40,8 +40,38 @@ npm install
...
-Currently unsupported.
-
+We offer a few ways of incorporating the GOV UK and HMRC CSS and JS libraries into your service. Pick the one which best suits your service.
+
+### 1. CDN
+If you are just using the GOVUK Frontend and HMRC Frontend and are not going to be adding any service-specific CSS or JS we suggest you use the HMRC CDN version of the assets.
+You will need to link to these in your service’s page template.
+
+1. add CDN js links to your `@bodyEnd` block in the layout template
+    - `<script type="text/javascript" src="https://www.tax.service.gov.uk/assets/hmrc-frontend/3.X.X/hmrc-frontend-3.X.X.min.js"></script>`
```
nataliecarey: Instead of `hmrc-frontend` I think this should be `govuk-frontend` as:

 - It's a more common use-case
 - The version 3.X.X seems to be referring to govuk-frontend

If you want to use `hmrc-frontend` it's best to change the version to `1.X.X`

It's probably worth explaining that `X` is a placeholder and linking to the list of versions in github.

vinnicombe: Good spot, 2 mins

src/components/timeout-dialog/_timeout-dialog.scss
```diff
@@ -1,7 +1,7 @@
-@import "node_modules/govuk-frontend/govuk/settings/all";
-@import "node_modules/govuk-frontend/govuk/tools/all";
-@import "node_modules/govuk-frontend/govuk/helpers/all";
-@import "node_modules/govuk-frontend/govuk/vendor/sass-mq";
+@import "../../../../govuk-frontend/govuk/settings/all";
+@import "../../../../govuk-frontend/govuk/tools/all";
+@import "../../../../govuk-frontend/govuk/helpers/all";
+@import "../../../../govuk-frontend/govuk/vendor/sass-mq";
```
nataliecarey, line 4: If we want to enforce this approach we can remove `node_modules` from the import list here https://github.com/hmrc/hmrc-frontend/blob/master/tasks/gulp/compile-assets.js#L56 and people using the old approach will lead to failed builds.

vinnicombe: Done

nataliecarey (approved):

PR #70, opened 2020-04-22
Remove compatibility check until mechanism can be fixed

matthewmascord:
The existing logic uses parseFloat to compare major and minor versions. E.g. there is a check that looks at parseFloat("1.10") = 1.1 < 1.10

### DIFF
--- a/check-compatibility.js
+++ b/check-compatibility.js
@@ -40,10 +40,15 @@ const checkCompatibility = (version, host = 'prototype-kit') => {
   const getMatchableVersion = (v) => parseFloat(v).toFixed(1)
 
   const installingHmrcVersion = String(getMatchableVersion(hmrcFrontendVersion))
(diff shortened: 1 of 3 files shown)


timsb (dismissed):

check-compatibility.js
```diff
@@ -40,10 +40,15 @@ const checkCompatibility = (version, host = 'prototype-kit') => {
   const getMatchableVersion = (v) => parseFloat(v).toFixed(1)
 
   const installingHmrcVersion = String(getMatchableVersion(hmrcFrontendVersion))
-  const latestKnownHost = compatibility[installingHmrcVersion][host][0]
+  const compatibilityMatch = compatibility[installingHmrcVersion]
+  if (!compatibilityMatch) {
+    console.log('The version couldn\'t be matched, compatibility couldn\'t be checked')
```
timsb: Maybe `console.error()` is better here?

timsb (approved):

PR #77, opened 2020-06-08
Standardising null and empty arrays

nataliecarey:
This is the HMRC equivalent of https://github.com/alphagov/govuk-frontend/pull/1638 which is best explained in this issue https://github.com/alphagov/govuk-frontend/issues/1618.

This PR includes an update of the `header` component which had fallen out of sync with `govuk-frontend`.  No other changes were required in the production code but I've added a test to make sure the output remains the same in these cases where `itemList` is used in `add-to-a-list`.

### DIFF
--- a/config/paths.json
+++ b/config/paths.json
@@ -13,6 +13,6 @@
   "packageTest": "__tests__/packageTest",
   "ports": {
     "app": 3000,
-    "test": 8888
+    "test": 8887
   }
 }
--- a/config/paths.json
+++ b/config/paths.json
@@ -13,6 +13,6 @@
   "packageTest": "__tests__/packageTest",
   "ports": {
     "app": 3000,
-    "test": 8887
+    "test": 8888
   }
 }
--- a/src/components/add-to-a-list/template.test.js
+++ b/src/components/add-to-a-list/template.test.js
@@ -171,4 +171,19 @@ describe('Add to a list', () => {
       expect($form.attr('action')).toBe('#addItem')
     })
   })
+
+  describe('Empty, null and missing itemLists', () => {
+    function overrideItemList (itemList) {
+      const params = {...examples.default}
+      params.itemList = itemList
+      return params
+    }
+
+    it('should have the same output for all versions of no items being provided', () => {
+      const outputs = [undefined, null, []].map(itemList => render('add-to-a-list', overrideItemList(itemList)).html())
+
+      expect(outputs[0]).toEqual(outputs[1])
+      expect(outputs[0]).toEqual(outputs[2])
+    })
+  })
 })
--- a/src/components/header/template.njk
+++ b/src/components/header/template.njk
@@ -2,12 +2,12 @@
 {% from "../banner/macro.njk" import hmrcBanner %}
 
 <header class="govuk-header hmrc-header {{ params.classes if params.classes }} {{ 'hmrc-header--with-additional-navigation' if params.signOutHref or params.languageToggle }}" role="banner" data-module="header"
-        {%- for attribute, value in params.attributes %} {{ attribute }}="{{ value }}"{% endfor %}>
+        {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
   <div class="govuk-header__container {{ params.containerClasses | default('govuk-width-container') }}">
     <div class="govuk-header__logo">
       <a href="{{ params.homepageUrl | default('/') }}" class="govuk-header__link govuk-header__link--homepage">
         <span class="govuk-header__logotype">
-          {# We use an inline SVG for the crown so that we can cascade the
+          {#- We use an inline SVG for the crown so that we can cascade the
           currentColor into the crown whilst continuing to support older browsers
           which do not support external SVGs without a Javascript polyfill. This
           adds approximately 1kb to every page load.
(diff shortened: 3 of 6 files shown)


src/components/header/template.njk
```diff
@@ -56,38 +55,29 @@
...
-            {% endfor %}
-          </ul>
-        </nav>
-      {% endif %}
+    {% if params.serviceName %}
+    <a href="{{ params.serviceUrl }}" class="govuk-header__link govuk-header__link--service-name">
+      {{ params.serviceName }}
+    </a>
+    {% endif %}
+    {% if params.navigation %}
```
matthewmascord: Isn't the idea of this change to not show the navigation if an empty list is provided?

config/paths.json
```diff
@@ -13,6 +13,6 @@
   "packageTest": "__tests__/packageTest",
   "ports": {
     "app": 3000,
-    "test": 8888
+    "test": 8887
```
timsb: Why does this port number need changing?

src/components/header/template.njk
```diff
@@ -56,38 +55,29 @@
         {% endif %}
       </a>
     </div>
-
-    {% if params.serviceName or params.navigation or params.signOutHref or params.languageToggle %}
+    {% if params.serviceName or params.navigation  %}
```
timsb: Isn't the `params.signOutHref` and `params.languageToggle` on of the reasons this was forked initially? Should we keep these in?

src/components/header/template.njk
timsb: Yeah, I would have thought the same. So should be:
```suggestion
    {% if params.navigation | length %}
```

src/components/add-to-a-list/template.test.js
```diff
@@ -171,4 +171,19 @@ describe('Add to a list', () => {
...
+    function overrideItemList (itemList) {
+      const params = {...examples.default}
+      params.itemList = itemList
+      return params
+    }
+
+    it('should have the same output for all versions of no items being provided', () => {
+      const outputs = [undefined, null, []].map(itemList => render('add-to-a-list', overrideItemList(itemList)).html())
+
+      expect(outputs[0]).toEqual(outputs[1])
```
timsb, line 185: Are there any other places that could potentially have this check? Not expecting it for this PR but just wondering for future improvements.

config/paths.json
nataliecarey: Doh.  I have conflicts every time I run the tests, I didn't mean to change it ... but if it's OK to change it I won't complain.   8888 is a common port, it might be worth moving away from it but unless that's immediately unanimous I'll revert this change.

src/components/add-to-a-list/template.test.js
nataliecarey: I was surprised to find that this is the only `type: array` that's documented (in the YAML) and I went through all the component sources to look for more - I can't find any.

It's a check worth adding any time a component accepts an array.

src/components/header/template.njk
nataliecarey: Hmm, I got this from GDS.  It's a decision between consistency and correctness I guess.  I'm happy to add it.

timsb: Is this not what the PR is for though? Just asking because I might be misinterpreting the reason for the PR. Should this be upstreamed if this is accepted? I think the navigation might have been changed slightly recently so may not be applicable anymore.

config/paths.json
timsb: That's fine. As you say 8887 is probably less likely to conflict so happy to leave it.

src/components/add-to-a-list/template.test.js
timsb: Ok, thanks for looking through! I'll keep an eye out for any future components.

src/components/header/template.njk
matthewmascord: +1 What's the reason for taking these out?

src/components/header/template.njk
```diff
@@ -56,38 +55,29 @@
         {% endif %}
       </a>
     </div>
-
-    {% if params.serviceName or params.navigation or params.signOutHref or params.languageToggle %}
+    {% if params.serviceName or params.navigation | length %}
```
timsb: Do we still need the `params.signOutHref` here?

src/components/header/template.njk
```diff
@@ -56,38 +55,29 @@
...
-                  {{ item.text }}
-                  </a>
-                </li>
-              {% endif %}
-            {% endfor %}
-          </ul>
-        </nav>
-      {% endif %}
+    {% if params.serviceName %}
+    <a href="{{ params.serviceUrl }}" class="govuk-header__link govuk-header__link--service-name">
```
timsb, line 61: There's a discussion going on at the moment about if the service name is displayed but serviceUrl isn't supplied. Not a change for this PR necessarily but wanted to mention.

nataliecarey: Yeah, it's a good conversation to have but I'd see it as separate to this PR.

src/components/header/template.njk
nataliecarey: Yes we do, I'll add that back in. Sorry.

nataliecarey:
Updated :)

timsb (approved):

matthewmascord:
Closing as superceded by https://github.com/hmrc/hmrc-frontend/pull/81

PR #83, opened 2020-06-29
PLATUI-454: Minor refactor for readability of timeout dialog and tests

nataliecarey:


nataliecarey:
I shouldn't really have raised a PR for this, I'll do it a different way.

PR #84, opened 2020-06-30
PLATUI-450: Resolve security vulnerabilities

matthewmascord:
Migrate to Gulp v4
Bump Node version to be consistent with govuk-frontend
Bump dependencies
Migrated to stylelint as sass-lint is unmaintained and has a vulnerability
Remove sassdoc (unused and has a vulnerability)


src/components/banner/template.test.js
```diff
@@ -9,9 +9,14 @@ const examples = getExamples('banner')
 describe('Internal Header', () => {
   describe('by default', () => {
     it('passes accessibility tests', async () => {
-      const $ = render('new-tab-link', examples.default)
+      const $ = render('internal-header', examples.default)
+
+      const results = await axe($.html(), {
+        rules: {
+          'region': { enabled: false }
```
nataliecarey, line 16: I feel that we should be looking to provide more life-like examples instead of turning off rules - maybe this can be done at a later time so it doesn't bloat the security update.

nataliecarey (approved):

PR #93, opened 2020-09-10
Timeout Dialog: allow users to set a different URL for timeouts than manual sign outs

gunjam:
Hi,

This pull request will allow users of the Timeout Dialog to set different url to redirect to on timeout than the URL for clicking the delete data / sign out link.

It should be none breaking, as if you don't set a timeout-url it will default to the sign-out-url, retaining the original functionality.

## Use case
Session timeouts are handled internally in our framework, so if you request a page after your session is timed out you are redirected to `/session-timeout` with a referrer querystring containing the URL of the page you requested (we use this in GA to see which pages people time out on).

Using this component I have to redirect to a `/end-session` URL to clear the session to makes the link work, but I lose the benefit of the querystring on the session timeout page. This is because it will also redirect to `/end-session`, then redirect to session-timeout with `/end-session` as the referrer.

Another issue is that since both the timeout and link will focibly kill the session, if the session is being refreshed in a separate tab, the fogotten tab will eventually timeout and kill a session for a user that wouldn't have otherwise reached it's expiry time.

However, if I set the timeout URL to redirect to the page it's on the session timeout middleware will kick in showing a timeout page with the approrpriate referrer URL and if the session is being maintained in another tab the page will just refresh every 30 minutes and not kill an active session. The sign out link however can continue to point to a URL that will forcibly end the session.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [1.16.0] - 2020-09-10
+
+### Changed
+
+- `hmrcTimeoutDialog` Allow users to set a different URL for timeouts than manual sign outs
+
 ## [1.15.3] - 2020-08-14
 
 ### Fixed
--- a/src/components/timeout-dialog/timeout-dialog.js
+++ b/src/components/timeout-dialog/timeout-dialog.js
@@ -58,6 +58,7 @@ function TimeoutDialog ($module) {
       countdown: validate.int(lookupData('data-countdown')),
       keepAliveUrl: validate.string(lookupData('data-keep-alive-url')),
       signOutUrl: validate.string(lookupData('data-sign-out-url')),
+      timeoutUrl: validate.string(lookupData('data-timeout-url')),
       title: validate.string(lookupData('data-title')),
       message: validate.string(lookupData('data-message')),
       messageSuffix: validate.string(lookupData('data-message-suffix')),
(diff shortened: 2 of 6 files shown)


ghost:
Can one of the admins verify this patch?

CHANGELOG.md
```diff
@@ -7,11 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
 ## [Unreleased]
 
-## [1.15.3] - 2020-08-14
+## [1.16.0] - 2020-09-10
 
-### Fixed
+### Changed
 
-- issue where internal header where font wasn't being set
+- `hmrcTimeoutDialog` Allow users to set a different URL for timeouts than manual sign outs
```
nataliecarey, line 14: This seems to be replacing the previous release rather than adding a new one, can you 

```
## [Unreleased]

## [1.16.0] - 2020-09-10

### Changed

- `hmrcTimeoutDialog` Allow users to set a different URL for timeouts than manual sign outs

## [1.15.3] - 2020-08-14

### Fixed

- issue where internal header where font wasn't being set

```

gunjam: Oops! Should be fixed now.

src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -255,7 +259,7 @@ function TimeoutDialog ($module) {
       var counter = getSecondsRemaining()
       updateCountdown(counter, $countdownElement)
       if (counter <= 0) {
-        signOut()
```
timsb, line 258: Looks like the signout function isn't used anymore. Can it be removed?

timsb (changes requested):

src/components/timeout-dialog/timeout-dialog.js
gunjam: `signOut()` is used for for the link and `timeout()` for the timeout, although since they aren't *re-used* I could just inline the Redirect call?

timsb (dismissed):

timsb (approved):

PR #104, opened 2020-10-21
Welsh language footer

nataliecarey:
We've had a number of requests for Welsh Language Footers, GDS aren't providing this so it's time to treat it seems appropriate to treat the Welsh Footer as an HMRC requirement.  This PR includes a clone of the current GOVUK footer with Welsh text added.

Welsh translations were taken from https://www.tax.service.gov.uk/pay which I'm told is accurate.

This follows the approach we've taken for the `govukHeader` which needed customising for HMRC's usage leading to the creation of `hmrcHeader`.

### DIFF
--- a/src/components/footer/template.njk
+++ b/src/components/footer/template.njk
@@ -1,0 +1,87 @@
+<footer class="govuk-footer {{ params.classes if params.classes }}" role="contentinfo"
+        {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+  <div class="govuk-width-container {{ params.containerClasses if params.containerClasses }}">
+    {% if params.navigation | length %}
+      <div class="govuk-footer__navigation">
+        {% for nav in params.navigation %}
+          <div class="govuk-footer__section">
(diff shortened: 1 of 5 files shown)


src/components/footer/template.njk
```diff
@@ -0,0 +1,88 @@
...
+            {% endif %}
+          </div>
+        {% endfor %}
+      </div>
+      <hr class="govuk-footer__section-break">
+    {% endif %}
+    <div class="govuk-footer__meta">
+      <div class="govuk-footer__meta-item govuk-footer__meta-item--grow">
+        {% if params.meta %}
+          <h2 class="govuk-visually-hidden">{{ params.meta.visuallyHiddenTitle | default("Support links") }}</h2>
```
matthewmascord: "Support links" also needs translating into Welsh. See: https://github.com/hmrc/play-ui/blob/b8d425cfa7bc54605c20a439de7cbe4c3f2e6597/src/main/resources/messages.cy#L24

matthewmascord (approved):
Tested locally, it looks good to me.

PR #108, opened 2020-11-11
PLATUI-792: Add configurable baseUrl and referrerUrl to hmrcReportTechnicalIssue.

nataliecarey:
The baseUrl allows us to configure local environments differently to production environments, the referrerUrl needs to be passed through by the service.  Usage of these are shown in the new examples.


src/components/report-technical-issue/report-technical-issue.yaml
```diff
@@ -2,11 +2,23 @@ params:
   - name: serviceCode
     type: string
     required: true
-    description: The URL-friendly service name which is passed to Contact Frontend
+    description: The URL-friendly service name which is passed to Contact Frontend.
   - name: language
     type: string
     required: false
     description: Either "en" for english or "cy" for welsh.
+  - name: classes
```
matthewmascord, line 10: Was this missing before? Should it be classList?

nataliecarey: You're right, it should be according to https://github.com/hmrc/hmrc-frontend/blob/master/src/components/report-technical-issue/template.njk#L4 but this is the only component which uses `classList` - the others all use `classes`.  It was me who wrote that previous version and I have no idea why I didn't follow the standard.

I'd suggest we remove `classList` and use `classes` instead to be consistent with the rest of the project and `govuk-frontend`.

src/components/report-technical-issue/template.njk
```diff
@@ -1,7 +1,7 @@
 {% set language %}{% if params.language == 'cy' %}cy{% else %}en{% endif %}{% endset %}
 
 <a lang="{{ language }}" hreflang="{{ language }}"
-  class="govuk-link {{ params.classList }}"
+  class="govuk-link {{ params.classes }}"
   target="_blank"
-  href="/contact/problem_reports_nonjs?newTab=true{% if params.serviceCode %}&amp;service={{ params.serviceCode }}{% endif %}"
+  href="{{ params.baseUrl }}/contact/problem_reports_nonjs?newTab=true{% if params.serviceCode %}&amp;service={{ params.serviceCode | urlencode }}{% endif %}{% if params.referrerUrl %}&amp;referrerUrl={{ params.referrerUrl | urlencode }}{% endif %}"
 >{% if params.language == 'cy' %}Help gyda'r dudalen hon (yn agor ffenestr neu dab newydd){% else %}Get help with this page (opens in a new tab){% endif %}</a>
```
timsb: The govuk design system suggest 'opens in new tab' rather than 'opens in a new tab' - https://design-system.service.gov.uk/styles/typography/#opening-links-in-a-new-tab. Could you update this whilst you're looking at this component?

src/components/report-technical-issue/report-technical-issue.yaml
matthewmascord: I'm OK with that.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "1.19.1",
+  "version": "1.19.2",
```
matthewmascord: Maybe make this a minor version update as we've changed the API.

src/components/report-technical-issue/report-technical-issue.yaml
nataliecarey: I've updated the template to follow the convention now.

src/components/report-technical-issue/template.njk
nataliecarey: Updated. 🙂

package.json
nataliecarey: Fair, updated.

matthewmascord: Don't forget check-compatibility.js :-)

timsb (dismissed):

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## [1.19.2] - 2020-11-12
```
matthewmascord: Version needs bumping to 1.20

matthewmascord (approved):

timsb (approved):

PR #111, opened 2020-12-02
PLATUI-847: Fix accessibility issue with hmrcHeader

matthewmascord:
Resolves accessibility issue with hmrcHeader when displayHmrcBanner is set to true.

Previously this component defined two top-level 'banner' landmarks failing automated Axe accessibility tests. On advice from the HMRC accessibility team, the HMRC banner has been moved inside HEADER so that screen-reader users read the banner as part of the overall header and can skip to the main content more easily.

Related unresolved GDS issues:
https://github.com/alphagov/govuk-frontend/issues/1604
https://github.com/alphagov/govuk-frontend/issues/1679

This change has been tested on the following devices/browsers:
https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices
June 2020

Windows 10, IE11
Windows 10, Edge 87
Windows 10, Firefox 83
Mac OS Catalina Safari 14
Mac OS Catalina Safari 14 with Voiceover
Mac OS Catalina Firefox 81
Mac OS Catalina Chrome 87
iPhone 8 Chrome
iPhone 8 Safari
Android 9, Samsung Galaxy S10, Chrome
Android 9, Samsung Galaxy S10, Samsung Internet

### DIFF
--- a/src/components/header/template.njk
+++ b/src/components/header/template.njk
@@ -1,110 +1,120 @@
 {% set language = params.language | default('en') %}
 {% from "../banner/macro.njk" import hmrcBanner %}
 
-<header class="govuk-header hmrc-header {{ params.classes if params.classes }} {{ 'hmrc-header--with-additional-navigation' if params.signOutHref or params.languageToggle }}" role="banner" data-module="header"
-        {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
-  <div class="govuk-header__container {{ params.containerClasses | default('govuk-width-container') }}">
-    <div class="govuk-header__logo">
-      <a href="{{ params.homepageUrl | default('/') }}" class="govuk-header__link govuk-header__link--homepage">
-        <span class="govuk-header__logotype">
-          {#- We use an inline SVG for the crown so that we can cascade the
-          currentColor into the crown whilst continuing to support older browsers
-          which do not support external SVGs without a Javascript polyfill. This
-          adds approximately 1kb to every page load.
+<header>
+  <div class="govuk-header hmrc-header {{ params.classes if params.classes }} {{ 'hmrc-header--with-additional-navigation' if params.signOutHref or params.languageToggle }}" data-module="header"
+          {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
+    <div class="govuk-header__container {{ params.containerClasses | default('govuk-width-container') }}">
+      <div class="govuk-header__logo">
+        <a href="{{ params.homepageUrl | default('/') }}" class="govuk-header__link govuk-header__link--homepage">
+          <span class="govuk-header__logotype">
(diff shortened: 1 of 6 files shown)


src/components/header/template.njk
```diff
@@ -1,110 +1,120 @@
...
-        {%- for attribute, value in params.attributes %} {{attribute}}="{{value}}"{% endfor %}>
-  <div class="govuk-header__container {{ params.containerClasses | default('govuk-width-container') }}">
-    <div class="govuk-header__logo">
-      <a href="{{ params.homepageUrl | default('/') }}" class="govuk-header__link govuk-header__link--homepage">
-        <span class="govuk-header__logotype">
-          {#- We use an inline SVG for the crown so that we can cascade the
-          currentColor into the crown whilst continuing to support older browsers
-          which do not support external SVGs without a Javascript polyfill. This
-          adds approximately 1kb to every page load.
+<header>
```
timsb: Is there a reason you've removed `role=banner`? I believe GOV.UK add these roles to elements for older browsers

src/components/header/template.njk
```diff
@@ -1,110 +1,120 @@
...
+          </nav>
+        {% endif %}
+      </div>
       {% endif %}
     </div>
-    {% endif %}
   </div>
+  {% if params.displayHmrcBanner %}
+    <div class="govuk-width-container">
+      <div class="hmrc-banner">
```
timsb: Shouldn't this use the component?

src/components/header/template.njk
```diff
@@ -1,110 +1,120 @@
...
+        <div class="hmrc-organisation-logo">
+          <p class="govuk-body-s">{% if params.language === 'cy' %}
```
timsb: You've used `language` and `==` elsewhere rather than `params.language` and `===`. Should be consistent.

timsb (changes requested):

src/components/header/template.njk
matthewmascord: The component has the role banner, which is not allowed within another banner element.

src/components/header/template.njk
matthewmascord: Are we not generally polyfilling HTML5 for IE8-10?

src/components/header/template.njk
matthewmascord: Good spot, this logic was copied in from hmrcBanner, will amend.

src/components/header/template.njk
matthewmascord: Added back in.

src/components/header/template.njk
timsb: Can the role be removed from the component? I can't see why we'd use the hmrc-banner in any other context? It means we've then got one consistent component

timsb (changes requested):

timsb (approved):

PR #113, opened 2020-12-09
PLATUI-792: Replace serviceCode with serviceId for consistency with c…

matthewmascord:
…ontact-frontend

### DIFF
--- a/check-compatibility.js
+++ b/check-compatibility.js
@@ -12,7 +12,7 @@ if (!knownPrototypeKitNames.includes(consumerPackageJson.name)) {
 }
 
 const compatibility = {
-  '1.22': {
+  '1.23': {
     'prototype-kit': ['9.8', '9.7', '9.6', '9.5', '9.4', '9.3', '9.2', '9.1', '9.0', '9.9', '9.10']
   },
   '0.6': {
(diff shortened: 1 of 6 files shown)


check-compatibility.js
```diff
@@ -12,7 +12,7 @@ if (!knownPrototypeKitNames.includes(consumerPackageJson.name)) {
 }
 
 const compatibility = {
-  '1.22': {
+  '1.23': {
     'prototype-kit': ['9.8', '9.7', '9.6', '9.5', '9.4', '9.3', '9.2', '9.1', '9.0', '9.9', '9.10']
```
timsb: The prototype kit if on version 9.11.2, could you add relevant versions

src/components/report-technical-issue/template.njk
```diff
@@ -1,7 +1,7 @@
 {% set language %}{% if params.language == 'cy' %}cy{% else %}en{% endif %}{% endset %}
-
+{% set serviceId = params.serviceId or params.serviceCode %}
```
timsb, line 2: Just wondering does Nunjucks evaluate this left to right? If servceId and serviceCode is passed then does it just pick up serviceId?

timsb (changes requested):

check-compatibility.js
matthewmascord: Thanks, updated now.

src/components/report-technical-issue/template.njk
matthewmascord: Good question. I've added a test for this edge case. serviceId takes precedence.

timsb (approved):

PR #118, opened 2020-12-16
Added ES Lint, following standards Plat-UI are using.  

nataliecarey:
I've fixed the issues highlighted and updated formatting to match expectations.

Most IDEs allow for eslint formatting on save, we recommend using it.  Many formatting issues can be resolved automatically, this can be achieved by running `npm run lint:fix`.  Linting will run alongside (after) the tests.


gulpfile.js
```diff
@@ -1,36 +1,31 @@
-'use strict'
```
timsb, line 1: Just out of interest, why are we removing 'use strict'?

matthewmascord: I think eslint has this rule because it is assuming everyone is using 'babel' that adds it for you. Do we know if rollup does this too? If not it might be safer to disable this rule as it does change the browser's behaviour.

matthewmascord (approved):

timsb (approved):

PR #123, opened 2021-02-17
PLATUI: Update govuk-frontend to v3.11.0

matthewmascord:


timsb (changes requested):
Can you update the hmrc-frontend version in package.json. Otherwise, looks good.

timsb (approved):

PR #134, opened 2021-03-25
Timeline

timsb:


src/components/timeline/README.md
```diff
@@ -0,0 +1 @@
+Timeline
```
matthewmascord: Is it worth having this README if empty?

src/components/timeline/_timeline.scss
```diff
@@ -0,0 +1,44 @@
+.hmrc-timeline {
```
matthewmascord, line 3:
```suggestion
@import "../../../../govuk-frontend/govuk/helpers/typography";

.hmrc-timeline {
```

matthewmascord (approved):

PR #137, opened 2021-04-21
Allowing multiple Character Count components on one page.

nataliecarey:
The problem report page currently has a broken character count (https://www.tax.service.gov.uk/contact/problem_reports_nonjs), the first one works and the second one doesn't. That's because `init-all.js` uses `querySelector` instad of `querySelectorAll` (https://github.com/hmrc/hmrc-frontend/blob/master/src/all.js#L22).

This can be reproduced in the prototype kit using this template:

```

{% extends "layout.html" %} {% block pageTitle %}
GOV.UK Prototype Kit

{% endblock %}
{% block content %} {%- from "hmrc/components/character-count/macro.njk" import hmrcCharacterCount -%}


{{ hmrcCharacterCount({
name: "abc",
id: "abc",
maxlength: 200,
label: { text: "This one will work", classes: "govuk-label--l" }
}) }}
{{ hmrcCharacterCount({
name: "def",
id: "def",
maxlength: 200,
label: { text: "This one will be broken", classes: "govuk-label--l" }
}) }}
{% endblock %}
```

Multiple character counts should be supported on one page.  This PR resolves the issue for anyone using `initAll`.

### DIFF
--- a/src/all.js
+++ b/src/all.js
@@ -2,6 +2,7 @@ import AccountMenu from './components/account-menu/account-menu';
 import TimeoutDialog from './components/timeout-dialog/timeout-dialog';
 import UserResearchBanner from './components/user-research-banner/user-research-banner';
 import CharacterCount from './components/character-count/character-count';
+import { nodeListForEach } from './common';
 
 function initAll() {
   const $AccountMenuSelector = '[data-module="hmrc-account-menu"]';
@@ -19,10 +20,10 @@ function initAll() {
     new UserResearchBanner($UserResearchBanner).init();
   }
 
-  const $CharacterCount = document.querySelector('[data-module="hmrc-character-count"]');
-  if ($CharacterCount) {
+  const $CharacterCounts = document.querySelectorAll('[data-module="hmrc-character-count"]');
+  nodeListForEach($CharacterCounts, ($CharacterCount) => {
     new CharacterCount($CharacterCount).init();
-  }
+  });
 }
 
 export default {
@@ -31,4 +32,5 @@ export default {
   TimeoutDialog,
   UserResearchBanner,
   CharacterCount,
+
 };
--- a/src/all.test.js
+++ b/src/all.test.js
@@ -1,0 +1,91 @@
+/* eslint-env jest */
+/* global spyOn */
+import HMRCFrontend from './all';
+import * as characterCountModule from './components/character-count/character-count';
+
+describe('Init All', () => {
+  let testScope;
+  beforeEach(() => {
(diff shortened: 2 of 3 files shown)


src/all.js
```diff
@@ -2,6 +2,7 @@ import AccountMenu from './components/account-menu/account-menu';
 import TimeoutDialog from './components/timeout-dialog/timeout-dialog';
 import UserResearchBanner from './components/user-research-banner/user-research-banner';
 import CharacterCount from './components/character-count/character-count';
+import { nodeListForEach } from './common';
 
 function initAll() {
```
matthewmascord, line 7: Is there a way we could write a test for this?

nataliecarey: For `nodeListForEach`?

matthewmascord: No the initAll function

nataliecarey: Yeah, that's fair.  It would be good to add those.

matthewmascord (dismissed):

src/all.test.js
```diff
@@ -0,0 +1,91 @@
...
+    };
+  });
+  afterEach(() => {
+    jest.resetAllMocks();
+    testScope.elemsToWipe.forEach((elem, key) => {
+      const parent = elem.parentNode;
+      if (parent) {
+        parent.removeChild(elem);
+      } else {
+        console.warn('failed to remove child', key);// eslint-disable-line no-console
```
timsb: Is this likely to happen? Just wondering why something like this would be in a test?

src/all.test.js
```diff
@@ -0,0 +1,91 @@
...
+
+    it('should not initialise when there are no modules', () => {
+      const initFn = jest.fn();
+      spyOn(characterCountModule, 'default').and.returnValue({ init: initFn });
+      HMRCFrontend.initAll();
+      expect(characterCountModule.default).not.toHaveBeenCalled();
+      expect(initFn).not.toHaveBeenCalled();
+    });
+
+    it('should initialise four module', () => {
```
timsb: Being really picky here but should be 'should initialise four modules'

src/all.test.js
nataliecarey: Fair, I hit that condition when I tried to remove the same item twice.  I think I might throw instead.

src/all.test.js
nataliecarey: Thank you.

nataliecarey: @timsb I've pushed a new version with these changes.

timsb (approved):

PR #138, opened 2021-04-22
Plat UI 1141 expand visual regression test coverage

oscarduignan:
Backstop is used for visual regression testing and previously we only compared the default example for each component.

In PLATUI-442 we're going to make a change to the add-to-a-list component.
That change wouldn't have been visible with our current backstop test coverage.

So before we do that, this change expands the backstop test coverage so that all examples are included.

It also moves the scenario specific backstop options into the component yaml files.

To try to make it easier to review there are two commits.
One updates the config for backstop and the second regenerates all the reference data.

Alongside those changes, this switch also uncovered a couple of bugs.


src/components/account-header/account-header.yaml
```diff
@@ -59,7 +59,6 @@ examples:
           href: '/components/account-header/with-service-name/preview'
         cy:
           href: '/components/account-header/welsh-with-service-name/preview'
-      serviceName: 'My service'
```
oscarduignan, line 62: This key was a duplicate

package.json
```diff
@@ -40,9 +40,9 @@
     "govuk-frontend": "^3.11.0"
   },
   "devDependencies": {
-    "autoprefixer": "^9.6.1",
     "@babel/preset-env": "^7.12.11",
-    "rollup-plugin-babel": "^4.4.0",
+    "ajv": "^6.12.2",
```
oscarduignan, line 44: ajv (json schema validation to check component config is correct) and glob (read files with a wildcard path) were already indirect dependencies so although they are new here they aren't new to the project (no update package lock needed)

app/app.js
```diff
@@ -103,7 +103,7 @@ module.exports = (options) => {
     const globalData = (type === 'layout' && exampleConfig.data) || {};
 
     if (!exampleConfig) {
-      next();
+      return next();
```
oscarduignan, line 106: for a component that had examples if you requested an example which didn't exist you'd get a misdirecting error logged rather than just not finding the example

src/components/account-header/account-header.yaml
nataliecarey: Thanks, I've been meaning to remove that for about a year :)

tasks/gulp/backstop-config.js
```diff
@@ -14,79 +28,59 @@ module.exports = ({ host, port }) => ({
...
-      url: `http://${host}:${port}/components/user-research-banner/preview`,
-      delay: 2000,
-    },
-    {
-      label: 'HMRC Character Count',
-      url: `http://${host}:${port}/components/character-count/preview`,
-      delay: 2000,
-    },
-  ],
+  scenarios: components.flatMap(({ componentsPath, componentsConfig }) => {
```
matthewmascord: As this logic is a bit more involved than previously, I wonder if it's worth splitting it out as a separate function with its own unit tests?

oscarduignan: yeah, I can pop it out and add some examples as tests 👍

src/components/timeout-dialog/timeout-dialog.yaml
```diff
@@ -30,3 +30,10 @@ examples:
     messageSuffix : "there is a suffix for the message"
     keepAliveButtonText: "keepAliveButtonText"
     signOutButtonText: "signOutButtonText"
+  visualRegressionTests:
+    backstopScenarioOptions:
+      delay: 10000
```
oscarduignan, line 35: Found that for this test nothing was displaying, it's happening for other timeouts too but I just updated this one because it's one that shows all the features of a timeout dialog so it's enough to get us coverage for most changes to the timeout dialog with other examples being mostly in welsh or around the timeout itself

tasks/gulp/backstop-config.js
```diff
@@ -14,79 +69,7 @@ module.exports = ({ host, port }) => ({
...
-      url: `http://${host}:${port}/components/user-research-banner/preview`,
-      delay: 2000,
-    },
-    {
-      label: 'HMRC Character Count',
-      url: `http://${host}:${port}/components/character-count/preview`,
-      delay: 2000,
-    },
-  ],
+  scenarios: buildScenarioList(host, port, components),
```
oscarduignan, line 72: Refactored this into it's own function and hopefully simplified it a bit. I didn't split it into a separate file because it will change at the same time / for the same reason as everything else in the file. My one todo here is to add a backstop-config.test.js which I'll pop in shortly

oscarduignan:
@natcarey / @matthewmascord I've updated the PR with some tests and a bit of refactoring, would you be able to review this again for me? I also added a commit to update the documentation

matthewmascord (approved):

PR #139, opened 2021-04-27
PLATUI-442: apply accessibility fix for add-to-a-list

oscarduignan:
Adam Liptrot found that when interacting with the list using JAWS list
items dialog that the subject of each list item AND the labels for the
actions available to operate on that subject would be announced
alltogether rather than just the label for the subject.

Source of change request:
https://github.com/hmrc/design-patterns/issues/31#issuecomment-799628620

### DIFF
--- a/src/components/add-to-a-list/template.test.js
+++ b/src/components/add-to-a-list/template.test.js
@@ -6,12 +6,12 @@ const examples = getExamples('add-to-a-list');
 
 const getTableItems = ($) => {
   const $rows = $('.hmrc-add-to-a-list__contents');
-  const $changeLinks = $('span.hmrc-add-to-a-list__change', $rows);
-  const $removeLinks = $('span.hmrc-add-to-a-list__remove', $rows);
+  const $changeLinks = $('.hmrc-add-to-a-list__change', $rows);
+  const $removeLinks = $('.hmrc-add-to-a-list__remove', $rows);
 
   const listItems = {};
   listItems.identifiers = [];
-  $('span.hmrc-add-to-a-list__identifier', $rows).each((index, element) => {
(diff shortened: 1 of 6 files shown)


src/components/add-to-a-list/template.test.js
```diff
@@ -23,6 +23,15 @@ const getTableItems = ($) => {
 };
 
 describe('Add to a list', () => {
+  it('should be rendered using a structure accessible for JAWS users', () => {
+    // See Adam Liptrot's comment on
```
matthewmascord: I would probably omit individuals' names as it's clear from the linked issue anyway.

oscarduignan: Good call, I'll update it!

oscarduignan:
failing PR is because I reverse patched a the change to remove some reformatting that my IDE applied and forgot about some changes to test selectors 🤦

oscarduignan:
@adamliptrot-oc I've implemented the accessibility fix you raised for the add-to-a-list component for jaws users (switching to a definition list) but I don't have access to jaws myself to retest the change. I've spun up a preview of the change with the new structure at https://deploy-preview-2--mystifying-feynman-5acf88.netlify.app/components/add-to-a-list/multiple-specific-items/preview.html if it's ok to ask, would you be able to recheck this for me to confirm it's resolved the issue in JAWS and been implemented correctly. I've tested it myself with Voice Over and run it through cross device tests for visual regressions and it's otherwise good to go I think

adamliptrot-oc:
@oscarduignan I can confirm this works as expected in JAWS 2021 with Chrome.
Screenshot:
Screenshot 2021-04-28 at 15 16 25

matthewmascord (approved):

PR #141, opened 2021-04-29
PLATUI-1133: Allowing multiple character counts on one page.

nataliecarey:


nataliecarey:
This was raised to test an alternative.  It has served its purpose and is now being closed.

PR #159, opened 2021-05-31
Adds nested elements to Page Heading pattern

philsherry:
## Problem

The current Page Heading pattern has two main issues:

1. Lack of nested `legend`/`label` elements.
2. Sub-heading needed improving.

When the HMRC Page Heading pattern was originally added to the pattern library, [making labels and legends headings](https://design-system.service.gov.uk/get-started/labels-legends-headings/) wasn’t a thing. As a result, most service teams have spun up their own slightly different versions to allow for the section sub-heading *above* the `h1`, but also allow for the `legend` being the first item inside the `fieldset`.

Some teams also had issues with the contextual *“This section is…”* content, which is helpful for screen reader users during larger service journeys. It’s possible this was down to a lack of documentation and understanding.

## Solution

This amended pattern adds the following:
- nested `legend` and `label` elements as page headings with section sub-headings
  - `label` can exist in the original component *(easy)*
  - `legend` is in the GovukFrontend `fieldset` component *(less easy)*
- improved context elements which make translations less painful

### `label` as heading

This adds the pre-existing `govukLabel` component in a conditional statement.

```jinja
{%- if params.label -%}
  {{ govukLabel({…}) }}
{%- else -%}
  <h1 class="govuk-heading-xl">
    {{ params.text }}
  </h1>
{%- endif -%}
```
🧑‍💻 `src/components/page-heading/template.njk`

```yml
- name: label-as-heading
  data:
    id: "label-id"
    label:
      classes: "govuk-label--xl"
      isPageHeading: true
      text: "What is your name?"
```
🧑‍💻 `src/components/page-heading/page-heading.yaml`
Label as heading
> *“Image: Label as heading.”*

### `label` as heading with section heading

`context` allows for one full sentence, rather than splitting it between visual and visually-hidden. This has benefits:
- it reads out in a smooth sentence, rather than forcing a screen reader user to key through, midway through the sentence
- it gives teams a bit more freedom, rather than forcing them into using *“This section is”* (which could always be changed in their `messages` file, but it was still brought up on occasion)
- translations of full sentences are usually better than trying to split individual phrases to fit with specific elements

`section` uses `aria-hidden` because it’s a purely visual element now that `context` is doing all the work.

```jinja
{%- if params.section -%}
  <p class="govuk-caption-xl hmrc-caption-xl">
    <span aria-hidden="true">{{ params.section }}</span>
    <span class="govuk-visually-hidden">{{ params.context }}</span>
  </p>
{%- endif -%}
```
🧑‍💻 `src/components/page-heading/template.njk`
```yml
- name: label-as-heading-with-section
  data:
    context: "This section is Personal details"
    id: "label-id"
    section: "Personal details"
    label:
      classes: "govuk-label--xl"
      isPageHeading: true
      text: "What is your name?"
```
🧑‍💻 `src/components/page-heading/page-heading.yaml`

Label as heading, with section
> *“Image: Label as heading, with section.”*

### `legend` as heading with section heading

Because the `legend` only exists in the [`govukFieldset`](https://design-system.service.gov.uk/components/fieldset/) component, we need to extend that here, instead of trying to crowbar it into [`hmrcPageHeading`](https://github.com/hmrc/design-patterns/issues/149).

```jinja
<section aria-labelledby="{{ params.labelledby }}">
  <header class="govuk-caption-xl hmrc-caption-xl">
    <span id="{{ params.labelledby }}" aria-hidden="true">
      {{ params.section }}
    </span>
    <span id="{{ params.describedBy }}" class="govuk-visually-hidden">
      {{ params.context }}
    </span>
  </header>
  {% call govukFieldset({…}) %}
  {% if caller %}
    {#- if statement allows usage of `call` to be optional -#}
    {{ caller() }}
  {% elseif params.html %}
    {{ params.html | safe }}
  {% endif %}
  {% endcall %}
</section>
```
🥷🏼 `src/components/fieldset/template.njk`
```yml
- name: default
  data:
    context: "This section is Personal details"
    contextId: "context-id"
    describedBy: "context-id"
    html: |
      <div class="govuk-body">
        <p>This is some content to put inside the fieldset</p>
      </div>
    labelledby: "short-id"
    legend:
      classes: "govuk-label--xl"
      isPageHeading: true
      text: "What is your name?"
    section: "Personal details"
```
🧑‍💻 `src/components/fieldset/fieldset.yaml`

Legend as heading, with section
> *“Image: Legend as heading, with section.”*


adamliptrot-oc:
I'm not sure about having a hidden block outside of the fieldset.
Can it not be done by sending a code block through to the legend parameter?

For example, construct a variable to contain the html for the heading and the caption
```
Heading copy <span class="govuk-caption-xl hmrc-caption-xl">Caption copy</span>
```

and send this through to the component as the label

```
label = HtmlContent(varName)
```

then send some additional classes through to the label
```
labelClasses = Some("hmrc-page-heading govuk-!-margin-top-0 govuk-!-margin-bottom-2")
```

timsb:
Sorry for the lack of reply to this – I realise it's been a long time with no comments. I've been talking with Plat-UI about this and we'd like to explore some of the other options that have been proposed to this as well as do some testing on your version.
We'd also like to look into Adam's comments above.

The HTML output from Martin Underhill's version in the Page heading issue (which originally came from you) looks simpler as it doesn't have the extra aria/roles attributes. Is there a reason you went with this markup for your version?

We'll look to see if there's any way we can simplify and make the component easier to maintain in the future but we're initially just going to explore different options for now and we'll let you know how we get on.

oscarduignan:
very late, but closing this because we updated the pattern https://design.tax.service.gov.uk/hmrc-design-patterns/page-heading/

PR #167, opened 2021-06-22
PLATUI-1212: propose two changes for add-to-a-list

oscarduignan:
- extract list-with-actions component from add-to-a-list component
- change api of add-to-a-list so it uses fixed form values for radios

meeting to make a decision for both still to be arranged

https://github.com/hmrc/design-patterns/issues/31

### DIFF
--- a/docs/adr/0003-change-api-of-add-to-a-list-form.md
+++ b/docs/adr/0003-change-api-of-add-to-a-list-form.md
@@ -1,0 +1,86 @@
+# Change API of add-to-a-list form
+
+* Status: accepted
+* Deciders: platui, design resources
+* Raised: 2021-06-22
+* Decided: 2021-07-01
+* Components: add-to-a-list
+
+Technical Story: PLATUI-1212
+
+## Context and Problem Statement
+
+The radio buttons in add-to-a-list are using the translated strings `messages.yes` and `messages.no` for their value.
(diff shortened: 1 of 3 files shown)


docs/adr/0003-change-api-of-add-to-a-list-form.md
```diff
@@ -0,0 +1,86 @@
...
+* we want to have a stable predictable interface for implementers to work to
+* reduce chance that the consequences of changes can be missed
+
+## Considered Options
+
+* Option 1: do nothing
+* Option 2: change to "true" and "false"
+* Option 3: change to "Yes" and "No" (capitalized)
+
+## Decision Outcome
```
JoPintoPaul, line 55: Do you think we should come down on a decision before publishing the ADR? Or make a note to amend this once it's decided? Thinking of future documentation value.

JoPintoPaul (dismissed):

docs/adr/0003-change-api-of-add-to-a-list-form.md
oscarduignan: I think it's probably a good idea to publish the decision now but I could change the status to "accepted, final implementation not yet decided" and have it as a criteria of the ticket to implement the change to update this. What do you think @JoPintoPaul?

matthewmascord:
Would it be worth deferring committing this until the implementation is actually done?

matthewmascord (approved):

PR #168, opened 2021-06-29
PLATUI-1256: Upgrade govuk-frontend to v3.13.0

matthewmascord:


package.json
```diff
@@ -42,7 +42,7 @@
   },
   "homepage": "https://github.com/hmrc/hmrc-frontend#readme",
   "dependencies": {
-    "govuk-frontend": "^3.12.0"
+    "govuk-frontend": "^3.13.0"
```
oscarduignan, line 45: @matthewmascord do you think we should remove the caret? Probably not really risky

oscarduignan (approved):

JoPintoPaul (approved):

PR #174, opened 2021-08-04
Dr 1105 timeout heading

timsb:
This PR adds default wording to the Service timeout dialog based on feedback from DAC and the DIAS team.

It effectively makes the heading non-optional but the wording of the heading is customisable for different service's needs.

After talking with Adam Liptrot it was decided that making the heading non-optional was the desired behaviour.
This PR also adds the heading to the ariaLabelledBy attribute of `#hmrc-timeout-dialog`


oscarduignan (approved):

src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -41,7 +41,7 @@ function TimeoutDialog($module) {
         second: 'eiliad',
       },
     } : {
-      title: undefined,
+      title: 'You’re about to be signed out',
       message: 'For your security, we will sign you out in',
```
adamliptrot-oc, line 45: Should we provide default messages for non-authenticated services too?
Leaving this to config options would mean inconsistency and is unneeded calls on the teams to provide translations etc.
Maybe default to authenticated but allow a swithc via a param like `isNotAuth` or similar to show the unauthenticated copy?

timsb: Yeah I considered this but as there were far fewer unauthenticated services I didn't add it as part of this PR. I'd be happy to add it as a further piece of work.

Had a look in Github and it seems 'Your application will time out soon' seems to be the most common way of timing out an authenticated session. This wording is problematic though as not all unauthenticated services will be an application so will need some thinking about how to word it (maybe more along the lines of 'Your session will time out soon').

Happy for me to raise it with the rest of the team for a further PR?

adamliptrot-oc: Yep, sounds good.

adamliptrot-oc (approved):

PR #175, opened 2021-08-05
PLATUI-1188: add links to examples under components on the listing page

oscarduignan:
Extracted from a previous change https://github.com/hmrc/hmrc-frontend/pull/140/files

listing-page-with-examples

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [2.1.1] - 2021-08-05
+
+### Changed
+
+- Added links to examples under components on the listing page when running the local dev server
+
 ## [2.1.0] - 2021-08-04
 
 ### Changed
--- a/app/views/layouts/index.njk
+++ b/app/views/layouts/index.njk
@@ -6,14 +6,18 @@
   </h1>
 
   <div class="govuk-grid-row">
-
     <div class="govuk-grid-column-one-half">
       <h2 class="govuk-heading-m">Components</h2>
-      <ul class="govuk-list">
-        {% for componentName in componentsDirectory | sort %}
-          <li><a href="/components/{{ componentName }}/preview" class="govuk-link">{{ componentName | replace("-", " ") | capitalize }}</a></li>
-        {% endfor %}
-      </ul>
+      <p class="govuk-body">Follow the links under sub headings below to view examples of the components.</p>
+      {% for componentName in componentsDirectory | sort %}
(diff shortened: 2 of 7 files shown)


oscarduignan:
format on save got me, I'll fix all the formatting changes 🤦

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [2.2.0] - 2021-08-05
```
oscarduignan: realised that I have to update the package version in the PR too for this repo, so I'll change this to a minor release

oscarduignan: Done!

app/views/layouts/index.njk
```diff
@@ -9,11 +9,18 @@
 
     <div class="govuk-grid-column-one-half">
       <h2 class="govuk-heading-m">Components</h2>
-      <ul class="govuk-list">
+      <dl class="govuk-list">
```
timsb: I'd be tempted to do away with the `dl` and structure of having the primary example as the definition term and the secondary examples as definition descriptions as it's creating a hierarchy that I'm not sure is there. I'd probably go for a heading for the component name (no anchor element) and then a list of all the examples. Although the first example is possibly going to be the most used I don't think it's necessarily the best example.

Alternatively if you do want to keep the `dl` I'd do away with the `ul` as a `dt` can have multiple `dd`s so I've just make all the other examples `dd`s

timsb (changes requested):

app/views/layouts/index.njk
oscarduignan: heading and list sounds good to me, I'll update it, thank you @timsb !

oscarduignan: @timsb I have updated with your suggestion and I saw there was also something else that was possibly partially implemented around displaying a component's readme that I figured I might as well implement at the same time

CleanShot 2021-08-09 at 12 39 19@2x

If you click view the readme it takes you to `/components/:component` and I've added in the needed code to render the readme at this route. From a comment it seemed like this was already an intention that hadn't been implemented. Small affordance but maybe helpful and it was quick to add! I can remove it if you think it's not useful 🤷‍♂️ here's the diff of the changes https://github.com/hmrc/hmrc-frontend/compare/7f8f6cc6d5acb0a48427ac6a546f21f86ab55412..a42c5852ec996f29c427098f08e79054ef845902

timsb: Yeah I'm happy to link to READMEs. I think in a lot of places they're pretty sparse but hopefully this will encourage us/me to expand on them and make them more useful.
The markup changes look good to me as well.

timsb (approved):

PR #179, opened 2021-08-31
PLATUI-1268: List with actions component

matthewmascord:


src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -0,0 +1,63 @@
...
+.hmrc-list-with-actions--long {
+  @include govuk-media-query($from: desktop) {
+    .hmrc-list-with-actions__name {
+      width: 50%;
+    }
+  }
+}
+
+.hmrc-list-with-actions__item {
+  display: flex;
```
matthewmascord, line 28: Using flexbox instead of a table layout due to the possibility of different numbers of actions.

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -0,0 +1,63 @@
...
+
+.hmrc-list-with-actions__actions {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
+  flex-wrap: wrap;
+
+  @include govuk-media-query($from: desktop) {
+    justify-content: flex-end;
+    flex-direction: row;
```
matthewmascord, line 44: On Desktop align actions horizontally, on mobile stack vertically.

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "hmrc-frontend",
   "version": "2.2.2",
```
timsb: You need to remove version 2.2.2

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+
+.hmrc-list-with-actions__action {
+  margin-bottom: govuk-spacing(2);
+  text-align: right;
+
+  @include govuk-media-query($from: desktop) {
+    margin-bottom: 0;
+    justify-content: flex-end;
```
timsb: Looks like this isn't supported in IE11 (https://caniuse.com/mdn-css_properties_justify-content_flex_context_start_end) but I can't see any visual differences when testing. Is this needed?

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -0,0 +1,63 @@
...
+}
+
+.hmrc-list-with-actions__actions {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
+  flex-wrap: wrap;
+
+  @include govuk-media-query($from: desktop) {
+    justify-content: flex-end;
```
timsb: Looks like this isn't supported in IE11 (https://caniuse.com/mdn-css_properties_justify-content_flex_context_start_end) but I can't see any visual differences when testing. Is this needed?

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -0,0 +1,63 @@
...
+}
+
+.hmrc-list-with-actions__name {
+  font-weight: normal;
+}
+
+.hmrc-list-with-actions__actions {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
```
timsb: Looks like this isn't supported in IE11 (https://caniuse.com/mdn-css_properties_justify-content_flex_context_start_end) but I can't see any visual differences when testing. Is this needed?

timsb (changes requested):
Commenting on this here as can't see a way to comment on rich diffs: When looking at the rich diffs, there seems to be added space between heading and list with actions apart from on mobiles where it there is less space. Any ideas how this is happening?

matthewmascord:
> Commenting on this here as can't see a way to comment on rich diffs: When looking at the rich diffs, there seems to be added space between heading and list with actions apart from on mobiles where it there is less space. Any ideas how this is happening?

I think it's due to the white-space semantics of an empty table being different to an empty div. We could potentially adjust for this if we think it's a problem. What do you think?

package.json
matthewmascord: Done.

src/components/list-with-actions/_list-with-actions.scss
matthewmascord: It only makes a difference in the extreme example when the actions wrap onto the next line. However, given we wouldn't encourage this pattern, I suggest we remove for now in the interests of simplicity.

src/components/list-with-actions/_list-with-actions.scss
matthewmascord: Removed as per comment above.

src/components/list-with-actions/_list-with-actions.scss
matthewmascord: Removed as per comment above.

timsb:
All looks good apart from the margin/padding issue. I'll take a look to see what's causing it. Elements should have the same whitespace between the heading and itself so I'd expect it to be the same.

timsb:
One thing I spotted was that for the List with actions, when there are no items it's still taking up whitespace and has some markup associated with it. Could you have an if statement to only produce markup when there are items. I think this is important with this pattern when used as part of the Add to list.

matthewmascord:
Added change to not render the DL element in the case of an empty list.

timsb (approved):

PR #190, opened 2021-11-01
PLATUI-1419: initial changes to match PTA

oscarduignan:
Implementing changes from https://github.com/hmrc/frontend-template-provider/commit/ad7bf2dbb1de37ed4e5986dff68805a151d14ea0

Questions:
- not sure what the right param to use for the "your profile" link is, so we have created a new one. But it might be that we instead want to re-use the personal details parameter.
  - confirmed with Muhammad from PTA over slack that he was happy with the API change here


src/components/account-menu/account-menu.yaml
```diff
@@ -1,5 +1,5 @@
 params:
-- name: accountHome
+- name: accounmtHoe
```
oscarduignan: oops!

JoPintoPaul (approved):

PR #193, opened 2021-11-09
Applying npm audit fixes

oscarduignan:
The difference in the screenshot is a small change to the text baseline because of backstop changing how it renders, not an actual regression. There are still issues to be resolved but this fixes the "safe" stuff, that we could do with `npm audit fix`


oscarduignan:
converted to a draft because in fixing some bits on another ticket we found some ways to be able to fix more on this one I think. We avoided bumping the jest version because it broke a load of tests, but we found the two settings we need to add to the jest config to fix stuff: https://github.com/hmrc/tracking-consent-frontend/blob/109ddc9998afebf3019a4e086f318c0051001069/jest.config.js#L6-L7

package.json
```diff
@@ -114,6 +113,8 @@
     "snapshotSerializers": [
       "jest-serializer-html"
     ],
+    "testEnvironment": "jsdom",
+    "testRunner": "jest-jasmine2",
```
oscarduignan, line 117: jest 27 changed default behaviour, this gets us the legacy behaviours our tests depend on.

src/components/timeout-dialog/dialog.test.js
```diff
@@ -1,5 +1,4 @@
 /* eslint-env jest */
-import mock from 'jest-mock';
```
oscarduignan, line 2: jest is available globally and has the mock functions in it, only need jest-mock if you're not using jest (I think)

src/components/timeout-dialog/timeout-dialog.test.js
```diff
@@ -80,23 +79,23 @@ describe('/components/timeout-dialog', () => {
...
+        closeDialog: jest.fn(),
+        setAriaLabelledBy: jest.fn(),
+        addCloseHandler: jest.fn().mockImplementation((fn) => {
           testScope.latestDialogCloseCallback = fn;
         }),
       };
       return testScope.latestDialogControl;
     });
-    jest.useFakeTimers();
+    jest.useFakeTimers('legacy');
```
oscarduignan, line 98: jest 27 changed default behaviour, this gets us the legacy behaviours our tests depend on (useFakeTimers mocks things like window.setTimeout)

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [3.1.1] - 2021-11-09
+
+### Changed
+
+- Updated dependencies to get security fixes after running `npm audit`. Still some remaining issues, but seems like they
```
JoPintoPaul: Given that this is a public repo, maybe best not to put in the official docs that there might be remaining security issues? Genuine question.

package.json
JoPintoPaul: @oscarduignan Does there need to be any work to rewrite the tests, based on the above? either for us, or the Design Resources team?

CHANGELOG.md
oscarduignan: @JoPintoPaul what about "remaining things currently flagged by npm audit have been judged low to no risk to our use case." I think in terms of security, we're probably exposing ourselves to less risk by disclosing that we've reviewed the remaining things and judged them not a risk

oscarduignan: I reckon it would be very hard to exploit most of the things raised by npm audit, so in my head it's mostly reputational risk of being able to have shown to have triaged them

JoPintoPaul: @oscarduignan How about:
Updated dependencies based on output of `npm audit` where relevant.

oscarduignan: @JoPintoPaul perfect! Updated 👍

JoPintoPaul (approved):

PR #194, opened 2021-11-16
Updated HMRC Banner to use SVG 

tomshaw1:


timsb (dismissed):

klaudiarybicka (dismissed):

src/components/banner/template.njk
```diff
@@ -1,5 +1,23 @@
...
+      <path d="M67.47,7.72a59.76,59.76,0,1,0,59.75,59.75A59.75,59.75,0,0,0,67.47,7.72m67.47,59.75A67.47,67.47,0,1,1,67.47,0a67.47,67.47,0,0,1,67.47,67.47" />
+    </svg>
+  <!--<![endif]-->
+  <!--[if IE 8]>
+    <img src="/assets/components/banner/images/hmrc_crest_18px.png">
```
oscarduignan: @tomshaw1 instead of a hardcoded src here we'd need an assetsPath param like the header uses:
- param defined in the component options https://github.com/hmrc/hmrc-frontend/blob/fe15941b5e73b1d427b1594916f9440a0864b87f/src/components/header/header.yaml#L6-L9
- used in template https://github.com/hmrc/hmrc-frontend/blob/fe15941b5e73b1d427b1594916f9440a0864b87f/src/components/header/template.njk#L48

src/components/banner/template.njk
```diff
@@ -1,5 +1,23 @@
 <div class="hmrc-banner">
   <div class="hmrc-organisation-logo">
+  <!--[if gt IE 8]><!-->
+    <svg id="f2a14da0-d308-46dc-ad18-2464344bd5d1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" width="26px" height="26px" viewBox="0 0 134.94 134.94">
```
oscarduignan: @tomshaw1 I wonder if we should include this via linking to the svg file here rather than having it defined inline. We've got the file in the PR separately but then duplicated here. If we have it separate then it can be cached by people's browsers, only 7kb, but that 7kb extra will be sent with every request rather than cached because the html won't be being cached by browsers in most cases

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [3.2.0] - 2021-11-17
+
+### Changed
+
+- Updated HMRC banner to use SVG logo and tweaked vertical alignment
```
oscarduignan, line 14: @tomshaw1 / @timsb with this I can't see the ticket I think so it might have been explained already there, but are we sure the vertical alignment needed tweaking? I ask this as devils advocate just because I was looking at the "swipe" view of the github rich diff and I couldn't see much obvious misalignment previously 🤷‍♂️ just me curious, not a blocker 

CleanShot 2021-11-17 at 16 04 45@2x

Jon-Rowe-HMRC (commented):
The layout is quite different to https://design.tax.service.gov.uk/examples/hmrc-banner/example/

crest

^ Overlaying the banners on top of one another:
- the left bar is a different height and width
- the tracking is different on the type

src/components/banner/template.njk
tomshaw1: Thanks Oscar. I've changed the path to reflect this

src/components/banner/template.njk
tomshaw1: Hey @oscarduignan, the reason we did it inline is that GOV.UK does the same, which Tim found here https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/header/template.njk#L8-L19
I agree that it shouldn't be duplicated, so I'll delete the .svg from the images file.

CHANGELOG.md
tomshaw1: The reason we changed it is we saw the difference between ours and the current banner, and noticed that the current banner does not have equal spacing between the top of the border and the top of the image, and bottom of the border and bottom of the image. Hope that makes sense

tomshaw1:
> The layout is quite different to https://design.tax.service.gov.uk/examples/hmrc-banner/example/
> 
> crest
> 
> ^ Overlaying the banners on top of one another:
> 
> * the left bar is a different height and width
> * the tracking is different on the type

Hi @Jon-Rowe-HMRC, I've updated the banner so that the border size is in line with the current version. Tim mentioned that there is a difference between the banner locally and the banner in the Design Resources as the font is ever so slightly different locally, which explains the tracking looking a bit off
Screenshot 2021-11-18 at 15 10 29
Here is the 2 compared, and you can see the text is lined up exactly, the only difference being the small change we made to the vertical alignment

Jon-Rowe-HMRC:
'I've updated the banner so that the border size is in line with the current version' Thanks Tom

timsb (dismissed):

klaudiarybicka (dismissed):

Jon-Rowe-HMRC (dismissed):

CHANGELOG.md
oscarduignan: yup that makes sense!

src/components/banner/template.njk
oscarduignan: @tomshaw1 on this one you've done exactly what I asked, but now I see it I think there's something we might want to adjust (apologies for not seeing this before!) At the moment we've got `assetsPath` on the header described as "The public path for the assets folder." but actually the default value is not that, it's the path to the images folder within the assets folder. We inherited this from govuk https://github.com/alphagov/govuk-frontend/blob/d86739ea1bd5a2e866209bba3863a5820224db78/src/govuk/components/header/template.njk#L36 but they are a bit inconsistent themselves https://github.com/alphagov/govuk-frontend/blob/d86739ea1bd5a2e866209bba3863a5820224db78/package/govuk/template.njk#L17 and I think in this case, really assetsPath or assetPath should follow the convention of the second example (template.njk) where it's just to the root /assets then everything that comes after that should be hard coded into the components template themselves. So here it would change to `default('/assets') }}/images/govuk-logotype-crown.png"`. Otherwise keeping all the values for this in sync will be quite challenging. For example if I change the assetPath for the template, then to adjust it I can't just pass the value for the template to the header/banner as well, I have to know what the additional suffixed path is and adjust the assetsPath I use for each component.

oscarduignan: @tomshaw1 on this one too, I think the assetsPath parameter needs to be added to the banner.yml file too, only language param in there at the moment https://github.com/hmrc/hmrc-frontend/blob/fe15941b5e73b1d427b1594916f9440a0864b87f/src/components/banner/banner.yaml#L1-L5

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "3.1.1",
+  "version": "3.2.0",
```
oscarduignan, line 3: @tomshaw1 I spotted one more small thing, when you change the version number in here you need to do an npm install I think so that the version in package-lock.json increments as well.

src/components/banner/template.njk
tomshaw1: @oscarduignan that makes total sense, I've made that change now too thanks

package.json
tomshaw1: @oscarduignan gotcha, done that now too thank you

oscarduignan (approved):
all looks good to me! 🙌

timsb (approved):

PR #195, opened 2021-11-26
Replacing the Internal Header SVG with an inline version

tomshaw1:
Internal Header SVG should be inline to be consistent with the HMRC banner, and to allow variable colours for high contrast browsers.


klaudiarybicka (dismissed):

src/components/internal-header/_internal-header.scss
```diff
@@ -27,14 +27,20 @@
 .hmrc-internal-header__logo-link {
   display: table;
 
-  &::before {
-    content: "";
-    display: table-cell;
-    width: 52px;
+  p {
```
timsb: Would be good if this wasn't an element selctor but a BEM one. I'd move it out of the containing selector and call it something like `.hmrc-internal-header__logo-text` but keep the name consistent with hmrc-banner if possible.

timsb (changes requested):

src/components/internal-header/_internal-header.scss
```diff
@@ -46,6 +46,12 @@
   }
 }
 
+.hmrc-internal-header__logo-text {
+  display: inline-block;
```
timsb: Have been looking at this locally and found an inconsitency between the margin between the text and logo (interestingly though, backstop is passing). There is more margin on this version which I think is due to the element having `display: inline-block;` and whitespace (Take a look at https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#spaces_in_between_inline_and_inline-block_elements).
Thinking about this further, we probably want this element as an inline element as we don't want the logo and text to be split over multiple lines.
So, if you change this element to inline (either via the CSS or changing it to a span element and getting rid of the CSS rule) then we'll fix both issues.

timsb (changes requested):

src/components/internal-header/_internal-header.scss
tomshaw1: That makes sense, I'll change the CSS to keep it inline with HMRC banner for now. and thanks for the link :)

timsb (dismissed):

timsb (approved):

PR #196, opened 2021-12-01
Adding colour fix for high contrast devices on HMRC Banner and Internal Header 

tomshaw1:


src/components/banner/template.njk
```diff
@@ -1,7 +1,7 @@
 <div class="hmrc-banner">
   <div class="hmrc-organisation-logo">
   <!--[if gt IE 8]><!-->
-    <svg id="f2a14da0-d308-46dc-ad18-2464344bd5d1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" width="26px" height="26px" viewBox="0 0 134.94 134.94" focusable="false" aria-hidden="true">
+    <svg id="f2a14da0-d308-46dc-ad18-2464344bd5d1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" width="26px" height="26px" viewBox="0 0 134.94 134.94" focusable="false" aria-hidden="true" fill="currentColor">
```
timsb: The fill attribute is a Presentation attribute (https://www.w3.org/TR/SVG/attindex.html#PresentationAttributes) which means we can set this in the CSS rather than as an attribute (https://www.w3.org/TR/SVG/styling.html#StylingUsingCSS) as it keeps all of these colour properties in one place.

src/components/internal-header/template.njk
```diff
@@ -2,7 +2,7 @@
     <div class="hmrc-internal-header__logo">
         <a href="{{ params.homepageUrl | default('/') }}" class="hmrc-internal-header__logo-link">
         <!--[if gt IE 8]><!-->
-            <svg id="f2a14da0-d308-46dc-ad18-2464344bd5d1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" width="52px" height="32px" viewBox="0 0 134.94 134.94" focusable="false" aria-hidden="true">
+            <svg id="f2a14da0-d308-46dc-ad18-2464344bd5d1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" width="52px" height="32px" viewBox="0 0 134.94 134.94" focusable="false" aria-hidden="true" fill="currentColor">
```
timsb: See comment in banner component about fill attribute. Same applies here

timsb (changes requested):

src/components/internal-header/_internal-header.scss
```diff
@@ -12,6 +12,13 @@
   }
 }
 
+@media (forced-colors: active) {
+  svg {
+    forced-color-adjust: none;
+    color: canvastext;
```
timsb: Had a look at this again this morning and because the internal banner icon is a child of a link this should be `linktext`

timsb (changes requested):

timsb (dismissed):

timsb (dismissed):

timsb (dismissed):

timsb (approved):

PR #198, opened 2022-01-11
Plat UI 1532 uplift govuk frontend

JoPintoPaul:


kyle-bowden (approved):

oscarduignan (approved):

src/components/header/template.njk
```diff
@@ -67,9 +67,9 @@
         </{{ serviceNameTag }}>
       {% endif %}
       {% if params.navigation | length %}
-      <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
-      <nav>
-        <ul id="navigation" class="govuk-header__navigation {{ params.navigationClasses if params.navigationClasses }}" aria-label="Top Level Navigation">
+      <nav class="govuk-header__navigation" aria-label="Top Level Navigation">
+        <button type="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
```
timsb, line 71: I've checked on both main and this branch and I can't see this button. This button should be visible on mobile so that the navigation can be shown/hidden. At the moment, the navigation is hidden on 768px and below and can't be revealed. This might be to do with the fact that we've forked the header to accomodate user research banner, hmrc banner etc.
Not sure whether this needs to be fixed as part of this work or should be put on the backlog.

timsb (changes requested):

src/components/header/template.njk
JoPintoPaul: @timsb As discussed, this seems to have be an issue since at least v3.0.0, which will need further investigation as to and when this stopped working, and why the visual regression tests didn't catch this. As discussed, I'll open a ticket to fix that as a separate, post-release bug ticket - in the meantime, if you could approve as agreed, I'd be very grateful. Bug ticket: PLATUI-1550.

timsb (approved):

src/components/header/template.njk
timsb: Thank @JoPintoPaul !

PR #203, opened 2022-01-27
Added welsh translation example to the page-heading section subheading

kyle-bowden:


src/components/page-heading/page-heading.yaml
```diff
@@ -6,3 +6,8 @@ examples:
   data:
     text: "What is your name?"
     section: "Personal details"
+- name: with-section-welsh
```
oscarduignan, line 23: @kyle-bowden just noticed for page-heading we're missing a description of the params in this file, here's an example of it for another component https://github.com/hmrc/hmrc-frontend/blob/f030a78945763b8320b493733b9f6a1fdfdf1829/src/components/banner/banner.yaml#L1-L9 would we be able to add that in at the same time?

oscarduignan (approved):

PR #206, opened 2022-01-27
Fixed missing language param in page-heading yaml

kyle-bowden:


src/components/page-heading/page-heading.yaml
```diff
@@ -15,13 +15,15 @@ params:
...
     text: "What is your name?"
 - name: with-section
   data:
+    language: "en"
     text: "What is your name?"
     section: "Personal details"
 - name: with-section-welsh
   data:
-    language: cy
+    language: "cy"
```
oscarduignan, line 27: @kyle-bowden maybe we can use `required: false` instead https://github.com/hmrc/hmrc-frontend/blob/9bdddb948b6d07676df9f7378e5c89f0da486097/src/components/report-technical-issue/report-technical-issue.yaml#L12 and set default to english in play-frontent-hmrc

oscarduignan: https://github.com/hmrc/play-frontend-hmrc/blob/main/src/main/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/reporttechnicalissue/ReportTechnicalIssue.scala#L25 and https://github.com/hmrc/play-frontend-hmrc/blob/main/src/test/scala/uk/gov/hmrc/hmrcfrontend/views/viewmodels/reporttechnicalissue/Generators.scala#L28 I think are how it would map to view model and generator

kyle-bowden:
No needed

PR #211, opened 2022-03-15
Patch some bits raised by npm audit

oscarduignan:
By updating these dependencies:
- update backstopjs from v5 to v6
- update node-sass@7.0.1
- update standard

I skipped node-fetch because using npm-force-resolutions wasn't working
on jenkins and it seemed pragmatic to not spend more time because we
don't seem to be at risk.


oscarduignan:
going to have to remove node-fetch and npm-force-resolutions because it's not working with our CI

src/components/header/template.njk
```diff
@@ -91,8 +91,8 @@
               <li class="govuk-header__navigation-item">
                 <a{% if language == 'en' %} aria-current="page"{% else %} class="govuk-header__link"{% endif %}
                   lang="en" hreflang="en" href="{{ params.languageToggle.en.href }}">English</a>
-              </li>
-              <li class="govuk-header__navigation-item">
+		&nbsp;
+              </li><li class="govuk-header__navigation-item">
```
oscarduignan: the styles for the list items are using display inline-block which means that the list is whitespace sensitive, this fixes an issue where you could see a little white underline like this: `English _  |  Cymraeg` it might be better to adjust the css but that feels like a bigger change. Fixed it with this change because I noticed it was showing up in a lot of screenshots updated by these changes (due to slight text rendering differences with new backstop version)

tasks/gulp/backstop-config.js
```diff
@@ -86,4 +86,9 @@ module.exports = ({ host, port, components }) => ({
   asyncCompareLimit: 50,
   debug: false,
   debugWindow: false,
+  misMatchThreshold: 0,
+  resembleOutputOptions: {
+    ignoreAntialiasing: true,
+    usePreciseMatching: true
+  }
```
oscarduignan: otherwise it seems lots of changes aren't big enough compared to the size of the screenshot to actually show up as a change. If you look at the updated backstop screenshots you'll find underlines, spacing changes, and even one case where our reference image was a serif font and the test image was a sans-serif font but it showed as no changes...

src/components/language-select/_language-select.scss
```diff
@@ -1,7 +1,5 @@
 .hmrc-language-select {
-  @include govuk-typography-responsive($size: 16);
-  @include govuk-link-common;
-  @include govuk-link-style-text;
+  @include govuk-font($size: 16);
```
oscarduignan, line 2: this class is used not on the language select links, but on the container, so link styles aren't needed, however we need govuk-font styles to replace them because the language select has spans for the current item - which have no govuk class. This does not change the current language names colour, if -I wonder- maybe the link styles were originally intended to make the current language look like a link 🤷‍♂️

timsb (approved):

PR #214, opened 2022-03-18
implement the hmrc list-with-actions component using a govuk summary-list

kyle-bowden:
This PR is to update the hmrc-frontend components specifically, once merged we can then update play-frontend-hmrc to be inline with chnages.


README.md
```diff
@@ -36,6 +36,29 @@ npm install
...
+env BACKSTOP_TEST_HOST={your local machines ip address} npm run test:backstop 
+```
+
+NOTE: If you run into a problem where some of your visual regression tests are failing due to unrelated changes, for instance on screens you may not have touched or updated. 
+You may want to delete the `node_modules` folder from the root of the project and then re-run command
+
+```shell
+npm install
+```
+
```
JoPintoPaul: @kyle-bowden I think there is already documentation for `backstop` in the `CONTRIBUTING.md`, maybe that should be updated rather than adding into the `README`:
https://github.com/hmrc/hmrc-frontend/blob/main/CONTRIBUTING.md#visual-regression-tests

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -4,22 +4,46 @@
...
+
+  @include govuk-media-query($from: desktop) {
+    .govuk-\!-summary-list-key-width {
+      width: 50% !important;
+    }
+
+    .govuk-\!-summary-list-actions-width {
+      width: 50% !important;
+    }
+  }
```
timsb: These already exist in the gov.uk design system so we can get rid of these completely and just use them directly in `src/components/list-with-actions/template.njk`

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -4,22 +4,46 @@
...
-.hmrc-list-with-actions--short {
+.hmrc-list-with-actions--long {
   @include govuk-media-query($from: desktop) {
+    .govuk-\!-summary-list-key-width {
+      width: 30% !important;
+    }
+
+    .govuk-\!-summary-list-actions-width {
+      width: 70% !important;
+    }
```
timsb: There aren't equivalent classes for this – the closest ones are `govuk-!-width-two-thirds`, `govuk-!-width-one-third`. GOV.UK look like they'd suggest adding custom override classes here (https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#custom-override-classes). In our case I think we'd want to add `.hmrc-\!-summary-list-key-width` & `.hmrc-\!-summary-list-actions-width`

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -4,22 +4,46 @@
...
+
+    // Kept for backward compatibility with add-to-a-list
     .hmrc-list-with-actions__name {
-      width: 30%;
+      width: 50%;
```
timsb, line 18: Again, we can remove this class and just use an override class from GOV.UK

timsb (changes requested):

src/components/list-with-actions/_list-with-actions.scss
kyle-bowden: We are not using the hmrc-list-with-actions__name css style in the template anymore but leaving it here for backward compatibility.

src/components/list-with-actions/_list-with-actions.scss
kyle-bowden: I have instead added `hmrc-summary-list__key` and `hmrc-summary-list__actions` because they don't override anything and they are new CSS keys

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -4,22 +4,42 @@
...
   @include govuk-font($size: 19);
+
+  @include govuk-media-query($from: desktop) {
+    .hmrc-summary-list__key {
+      width: 50% !important;
+    }
+
+    .hmrc-summary-list__actions {
+      width: 50% !important;
+    }
```
timsb: Ah, just realised you can't use override classes for these (in the template) as they're in a media-query so keeping these around is fine 👍🏻

timsb (dismissed):

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [4.8.0] - 2022-03-23
```
oscarduignan: @kyle-bowden apologies, merged another change ahead of this so needs rebasing / incrementing version number

oscarduignan: and, because I've made a change to the backstop config you probably need to remove your backstop image changes then regenerate them

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -4,22 +4,42 @@
 
 .hmrc-list-with-actions {
   @include govuk-font($size: 19);
+
+  @include govuk-media-query($from: desktop) {
+    .hmrc-summary-list__key {
```
oscarduignan, line 9: @timsb @kyle-bowden why can't we target the .govuk-summary-list__key here? I'm not sure I understand the reason for introducing a new class when these are scoped to .hmrc-list-with-actions

src/components/list-with-actions/_list-with-actions.scss
```diff
@@ -4,22 +4,42 @@
...
-.hmrc-list-with-actions--short {
+.hmrc-list-with-actions--long {
   @include govuk-media-query($from: desktop) {
+    .hmrc-summary-list__key {
+      width: 30% !important;
+    }
+
+    .hmrc-summary-list__actions {
+      width: 70% !important;
+    }
```
oscarduignan: @kyle-bowden should 21 to 27 move into .hmrc-list-with-actions--short instead of here (.hmrc-list-with-actions--long)? Or are we deliberately changing the meaning here between implementations, so it was short/long name before, now it's short/long actions? If we're changing meanings, I'd almost opt for a different modifier if that's the case, and then we could make it clear what the modifier does, for example "--short-actions" or even "--30/70" "--50/50" cc @timsb

src/components/list-with-actions/_list-with-actions.scss
timsb: My thoughts were that if `.govuk-summary-list__key` were to change it would either be more maintenance for us or could introduce a change that goes unnoticed (admittedly that's less likely with VRTs)

oscarduignan (dismissed):

src/components/list-with-actions/_list-with-actions.scss
oscarduignan: so working it through, if govuk-summary-list markup was updated and someone saw that it was being used in this component then they might come and update it, but not realise that it was being used by css selector and then you could get a regression. That makes sense to me, and I can see how that could get past VRT tests, because person would already be expecting a change, so might not see that as well as the intended change there was a regression

oscarduignan: I wonder if we had it as hmrc-summary-list__key if we could maybe add an additional class to the container, alongside hmrc-list-with-actions we could have `hmrc-summary-list--without-value-column` so there's the old list with actions component, but also a more generic way that's just a hmrc variation of a summary summary list

timsb: I generally prefer separating our own styling from any of the govuk styling to keep the concerns separate. `.hmrc-summary-list--without-value-column` would be the way to go about this but I'm not sure I like the fact that teams will have to add this when they don't want the value column. Would the logic for this be in play-frontend-hmrc? Also, would it make more sense for the logic to be the other way around i.e. `.hmrc-summary-list--with-value-column`?

timsb (approved):

PR #215, opened 2022-03-23
Remove excess margin from above and below hmrcPageHeading component

kyle-bowden:


oscarduignan:
@kyle-bowden I think the package.json and package-lock.json need their version number bumping

src/components/page-heading/_page-heading.scss
```diff
@@ -1,18 +1,21 @@
 .hmrc-page-heading {
   display: table;
   width: 100%;
-  margin-top: govuk-em(60, 32);
-  margin-bottom: govuk-em(30, 32);
   padding-top: 8px;
   clear: both;
   table-layout: fixed;
 
+  [class^="govuk-caption"] {
```
oscarduignan: @kyle-bowden I think this should be ~ because ^ matches the start of the string, where if someone had implemented the markup themselves it might not always be at the start. I wonder if for this we should have, that way if someone is using their own type of styles for caption, they could use hmrc-caption to get the position change without the govuk-caption styles

```
[class~="govuk-caption"],
[class~="hmrc-caption"] {
```

oscarduignan: https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors actually I'm wrong about how ~= works, I think we want *=

oscarduignan: ~= would match `class="govuk-caption another-class"` but not `class="govuk-caption-xl another-class"`

kyle-bowden: I tried with ~ but it doesn't match for `govuk-caption-xl`

kyle-bowden: I have updated to use *= and its working as expected

kyle-bowden:
> @kyle-bowden I think the package.json and package-lock.json need their version number bumping

I have bumped the version now

src/components/page-heading/_page-heading.scss
```diff
@@ -1,18 +1,21 @@
 .hmrc-page-heading {
   display: table;
   width: 100%;
-  margin-top: govuk-em(60, 32);
-  margin-bottom: govuk-em(30, 32);
   padding-top: 8px;
   clear: both;
   table-layout: fixed;
 
+  [class*="govuk-caption"] {
```
timsb: Is there a reason we need to swap the class selector with an attribute selector? This raises the specificity value which is generally not a good thing. I'd prefer if we kept with BEM principles to keep specificity as low as possible. Could we keep the `.hmrc-caption-xl` – maybe rename to `.hmrc-caption` – and add `.hmrc-caption` and `.govuk-caption-xl` to the nunjucks.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "4.9.0",
+  "version": "5.0.0",
```
timsb: Is this a major version change now? I thought we'd decided this was just a minor version change because we're providing a default.

timsb: Is it due to the introduction of `.hmrc-page-heading--with-margin`? I'm not sure if just adding this new class needs a major version bump? Happy to discuss as I'm not 100% sold either way.

timsb (changes requested):

package.json
kyle-bowden: I have changed the versioning from `5.0.0` to `4.9.1` as this is only providing a default and adding a new style `.hmrc-page-heading--with-margin` shouldn't warrant a major version bump.

src/components/page-heading/_page-heading.scss
kyle-bowden: I have updated the styling in accordance with your comment on specificity and in keeping with BEM principles.

src/components/page-heading/template.njk
```diff
@@ -1,6 +1,6 @@
-<header class="hmrc-page-heading">
-  <h1 class="govuk-heading-xl">{{ params.text }}</h1>
+<header class="hmrc-page-heading{%- if params.classes %} {{ params.classes }}{%- endif -%}">
+  <h1 class="{{ params.headingClasses | default('govuk-heading-xl') }}">{{ params.text }}</h1>
   {%- if params.section -%}
-    <p class="govuk-caption-xl hmrc-caption-xl"><span class="govuk-visually-hidden">{% if params.language === 'cy' %}Teitl yr adran hon yw{%else%}This section is{%endif%} </span>{{ params.section }}</p>
+    <p class="{{ params.captionClasses | default('govuk-caption-xl hmrc-caption') }}"><span class="govuk-visually-hidden">{% if params.language === 'cy' %}Teitl yr adran hon yw{%else%}This section is{%endif%} </span>{{ params.section }}</p>
```
timsb: I think the `.hmrc-caption` should be outside of the conditional as we always want it to display as a table-header-group.

timsb (changes requested):

src/components/page-heading/template.njk
kyle-bowden: I have updated now, thanks for spotting that

timsb (approved):

PR #216, opened 2022-03-24
PLATUI-1646: add extra VRT tests for account header

oscarduignan:
Added a new helper for performing some steps before taking a VRT screenshot which was needed to capture some link states for the account menu before we fix some accessibility issues with their colour contrast

Added extra examples / VRT captures to the account header so we have coverage for focus and hover states of links before we make the change to increase the colour contrast of menu items

I've deliberately not bumped the version number because this is just some additional examples and test cases with no code changes


kyle-bowden (approved):

backstop_data/engine_scripts/puppet/beforeTakingScreenshotHelper.js
```diff
@@ -0,0 +1,28 @@
...
+  allowedPageActions.click = async selector =>
+    await page.click(selector);
+
+  allowedPageActions.hover = async selector =>
+    await page.hover(selector);
+
+  allowedPageActions.focus = async selector =>
+    await page.focus(selector);
+
+  allowedPageActions.type = async ({ into, text }) =>
```
kyle-bowden, line 13: I noticed .type, .press and .waitFor async functions are not used but I am happy for them to remain here.

PR #217, opened 2022-03-25
PLATUI-1646: add examples for hmrcAccountMenu

oscarduignan:
Needed a couple of additional ones for dropdown menu link states for when one is active.


src/components/account-header/account-header.yaml
```diff
@@ -47,10 +47,6 @@ examples:
         messageCount: '30'
       checkProgress:
         href: '/checkProgress'
-      paperlessSettings:
-        href: '/paperless/settings'
-      personalDetails:
-        href: '/personalDetails'
```
oscarduignan, line 53: these were no longer needed, previous change removed these options

kyle-bowden (approved):

PR #228, opened 2022-05-16
PLATUI-1691: record decision to use better-npm-audit in an ADR

oscarduignan:


docs/adr/0005-use-better-npm-audit.md
```diff
@@ -0,0 +1,77 @@
...
+### Problem statement
+
+Over time the number of issues we aren't able to fix has increased, and resulted in our dashboard showing all checks in red all the time. This makes it hard to notice new security vulnerabilities and means we spend time rechecking previous vulnerabilities.
+
+We would like the dashboard entries to not be displayed in red because of known issues that have already been reviewed, so that we can easily see where there are new vulnerabilities and avoid wasting time re-checking the same issues.
+
+## Decision Drivers
+
+* we would like to make a small change, so keep our current process, but improve the mechanism
+* we only want to check nodejs dependencies at the moment
```
JoPintoPaul, line 31: Is it worth adding in here that there are some options that are not currently supported by our projects, such as Dependabot?

JoPintoPaul (dismissed):

JoPintoPaul (approved):

PR #229, opened 2022-05-17
PLATUI-1721: Updated user-research-banner to have h2 heading

JoPintoPaul:


src/components/user-research-banner/template.njk
```diff
@@ -1,13 +1,13 @@
 <div class="hmrc-user-research-banner" data-module="hmrc-user-research-banner">
   <div class="hmrc-user-research-banner__container govuk-width-container">
     <div class="hmrc-user-research-banner__text">
-        <div class="hmrc-user-research-banner__title govuk-!-font-weight-bold">
+        <h2 class="hmrc-user-research-banner__title govuk-!-font-weight-bold">
```
oscarduignan: looking at the styles, we might be able to drop the `govuk-!-font-weight-bold` and use `govuk-heading-s` https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/core/_typography.scss#L47-L55 we might not need &__title { ... } styles at all then

timsb (approved):

PR #230, opened 2022-05-19
PLATUI-1723: update user research banner copy

oscarduignan:
Added extra visually-hidden text for screen readers

https://github.com/hmrc/platui-support/issues/15

### DIFF
--- a/src/components/user-research-banner/template.njk
+++ b/src/components/user-research-banner/template.njk
@@ -18,9 +18,11 @@
     </div>
     <button class="govuk-link hmrc-user-research-banner__close">
       {% if params.language == 'cy' %}
-        Dim diolch
+        <span aria-hidden="true">Dim diolch</span>{#- strip spaces -#}
+        <span class="govuk-visually-hidden">Dim diolch, nid wyf am gymryd rhan yn yr ymchwil defnyddwyr - cuddio'r neges hon</span>
       {% else %}
-        No thanks
+        <span aria-hidden="true">No thanks</span>{#- strip spaces -#}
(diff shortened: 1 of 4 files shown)


src/components/user-research-banner/template.njk
```diff
@@ -18,9 +18,11 @@
     </div>
     <button class="govuk-link hmrc-user-research-banner__close">
       {% if params.language == 'cy' %}
-        Dim diolch
+        <span aria-hidden="true">Dim diolch</span>{#- strip spaces -#}
+        <span class="govuk-visually-hidden">Dim diolch, nid wyf am gymryd rhan yn yr ymchwil defnyddwyr - cuddio'r neges hon</span>
```
ellamdav, line 22: is there a reason the Welsh version has a hyphen instead of a comma?  do they both cause the same behaviour for screenreaders (presumably a slight pause)?

oscarduignan: I'm not sure how commas and hyphens work with welsh, or I guess specifically a welsh screen reader, guess it would be different pronunciation rules. I've sent a screenshot of the translation in place with hyphen visible for them to check if I did it correctly, I'll send another note to check explicitly if the hyphen is right

oscarduignan: @ellamdav they've come back and said it was deliberate, but that we could change it if needed back to a comma. I'm going to ask in int-plat-design if Shabana knows of any issues with hyphens and screen readers

oscarduignan: Shabana has said that we should go with the recommended translation, rather than trying to accomodate screen readers - so this change is ready now 👍

kyle-bowden (approved):

PR #233, opened 2022-07-05
PLATUI-1874: Updated govuk-frontend version to v4.2.0

kyle-bowden:


JoPintoPaul:
Hi @kyle-bowden - I see in the v4.2.0 release notes that there is a change to the header, replacing `govuk-header__link--service-name` with `govuk-header__service-name`. 

We have a bespoke header within hmrc-frontend that references `govuk-header__link--service-name` - should we update the references within hmrc-frontend?

JoPintoPaul (approved):

PR #236, opened 2022-08-09
PLATUI-1896: sync session timeout dialogs

oscarduignan:
seems to work pretty well to prove we could use broadcast channel as a mechanism for this, however we don't check what session the activity came from so whichever page timed out first will "win" and if the next page shown to the user has a timeout dialog on it then it will reset the timeout of the other pages even though they were technically a part of the session that timed out. Issue might be most impactful if a background tab had a short timeout, it could keep the foreground page from ever displaying the timeout dialog - creates a slow infinite loop to if the timeout page includes a timeout dialog. Might not be an issue in practice, in any case I think maybe it's worth platform providing a standard timeout page that has bit better semantics than what services have- like maybe preserving CSRF token, so we just lose the auth session when user times out. Not something to look at for this ticket I think.

### DIFF
--- a/src/components/timeout-dialog/timeout-dialog.js
+++ b/src/components/timeout-dialog/timeout-dialog.js
@@ -15,6 +15,7 @@ function TimeoutDialog($module) {
   let settings = {};
   const cleanupFunctions = [];
   let currentTimer;
+  const sessionActivityChannel = window.BroadcastChannel && new window.BroadcastChannel('hmrc-timeout-dialog');
 
   cleanupFunctions.push(() => {
     if (currentTimer) {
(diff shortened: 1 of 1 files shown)


src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -76,8 +77,26 @@ function TimeoutDialog($module) {
...
+      sessionActivityChannel.postMessage(getDateNow());
+    }
+  };
+
+  const listenForSessionActivityAndResetDialogTimer = () => {
+    if (sessionActivityChannel) {
+      sessionActivityChannel.onmessage = (event) => {
+        const timeOfActivity = event.data;
+        cleanup();
+        setupDialogTimer(timeOfActivity);
```
oscarduignan, line 95: passing the timestamp from the fg to the bg tab in case the bg tab for some reason doesn't immediately get this message, maybe because the tabs js is suspended somehow

PR #242, opened 2022-08-24
PLATUI-1951: Update hmrc-header with latest changes from govuk-header

kyle-bowden:


src/components/header/template.test.js
```diff
@@ -232,11 +239,4 @@ describe('header', () => {
       expect($('.hmrc-banner > .hmrc-organisation-logo > p.govuk-body-s').text().trim()).toEqual('Cyllid a Thollau EM');
     });
   });
-
-  describe('additional banners block', () => {
```
oscarduignan: @kyle-bowden is this removal an unrelated change?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "5.4.0",
+  "version": "5.7.0",
```
oscarduignan: @kyle-bowden needs an `npm install` so package-lock.json gets updated too - but just to check, it's showing a jump of three versions - shouldn't it say 5.5.0? Latest released version is 5.5.0 - though I can see that changelog has 5.6.0 - not sure what's happened here with versions, might want to investigate this before we merge

oscarduignan: I guess this is the "docs changes don't bump versions" stuff maybe, even though we define the version explicitly I guess it's still short circuiting

src/components/header/template.test.js
kyle-bowden: @oscarduignan thanks for spotting this, this was removed accidentally.

oscarduignan (approved):

PR #243, opened 2022-09-06
PLATUI-1939: Update hmrc-frontend timeout dialogs so they can synchronise their timeouts using BroadcastChannel

kyle-bowden:


src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -267,8 +283,8 @@ function TimeoutDialog($module) {
   const keepAliveAndClose = () => {
     cleanup();
     setupDialogTimer();
-    utils.ajaxGet(settings.keepAliveUrl, () => {
-    });
+    broadcastSessionActivity();
+    utils.ajaxGet(settings.keepAliveUrl, () => {});
```
oscarduignan, line 286: @kyle-bowden / @ellamdav should broadcastSessionActivity following the keep alive request? The timeout dialog doesn't actually handle the keep alive request failing, so might not make a difference in practice

src/components/timeout-dialog/validate-input.js
```diff
@@ -10,4 +10,6 @@ ValidateInput.int = (stringToValidate) => {
 
 ValidateInput.string = (stringToValidate) => stringToValidate && ((`${stringToValidate}`) || undefined);
```
oscarduignan: total tangent, but I'm really curious about this ```(`${stringToValidate}`) || undefined```

src/components/timeout-dialog/session-activity-service.test.js
```diff
@@ -0,0 +1,71 @@
...
+
+    it('should call the passed callback fn with the data from the broadcast channel callback', () => {
+      const sut = new SessionActivityService(mockBroadcastChannelFactory);
+      const clientCallback = jest.fn();
+
+      sut.onActivity(clientCallback);
+
+      const broadcastChannelCallback = mockBroadcastChannel.onmessage;
+      const someTimestamp = Date.now();
+      broadcastChannelCallback({ data: { timestamp: someTimestamp } });
```
oscarduignan, line 59: @kyle-bowden / @ellamdav should this use `sut.logActivity()`?

oscarduignan:
@kyle-bowden / @ellamdav it looks good to me, few small questions but only bit missing I think is package.json / package-lock.json version bump. Guess docs make better sense in the play-frontend-hmrc readme

src/components/timeout-dialog/timeout-dialog.js
ellamdav: yeah, we considered that, but decided to decouple the two because there's no existing error handling, and they seemed like independent fire-and-forget events.  also, it would be more complicated to test the callback from `ajaxGet`.  do you think we need to consider what should happen if the ajax keepalive call were to fail?

oscarduignan: I think the timeout dialog should probably handle it, but not something worth adding in this change, just a retry would go some distance - or showing an error - at the moment it would act like it kept the session alive but if it failed (in certain ways) then it wouldn't have actually extended the session and the next time user tried to do something they might get an error. It's the kind of thing that I think would be much better defined as an explicit state machine https://xstate.js.org/docs/ because I bet there's a load of invalid states it can be in at the moment that haven't been thought about

src/components/timeout-dialog/session-activity-service.test.js
ellamdav: It _could_, we just thought it was better to explicitly show what was happening under the hood (ie. the unwrapping of the broadcastChannel `data`).

src/components/timeout-dialog/validate-input.js
kyle-bowden: It is not very readable, I'm still struggling to understand it, I would prefer something like

`
ValidateInput.string = (stringToValidate) => (typeof stringToValidate === 'string' ? stringToValidate : undefined);
`

oscarduignan (approved):

PR #247, opened 2022-09-15
PLAT-1959: Update report technical issue component to have an optional rel attribute based on the presence of a referrer url

kyle-bowden:


src/components/report-technical-issue/template.test.js
```diff
@@ -118,4 +111,11 @@ describe('Report Technical Issue', () => {
 
     expect($('.govuk-link').attr('href')).toEqual('/contact/report-technical-problem?newTab=true&service=pay&referrerUrl=http%3A%2F%2Flocalhost%3A9123%2Fmy-service');
   });
+
+  it('specify rel attributes as "noreferrer noopener" to prevent reverse tabnapping vulnerability when referrer url is provided', () => {
```
oscarduignan: @kyle-bowden should we have a test `it does not include a rel="noreferrer noopener" attribute if the referrerUrl is not explicitly passed, so that contact frontend can infer it from the referrer header`

oscarduignan (approved):

PR #249, opened 2022-09-21
PLATUI-1829: Added dependency on accessible-autocomplete and included…

JoPintoPaul:
… CSS and JS in build artefacts


tasks/gulp/webjar-helpers.js
```diff
@@ -87,13 +87,18 @@ const webJarHelpers = ({
...
+        const sanitisedDependencyVersion = dependencyVersion.replace(/[~^]/, '');
+        return `<dependency>
             <groupId>org.webjars.npm</groupId>
             <artifactId>${name}</artifactId>
             <version>${sanitisedDependencyVersion}</version>
         </dependency>`;
+      };
+
+      const entriesAsDependencies = Object.entries(dependencies).map(entryToDependency);
+      return entriesAsDependencies.toString().split(',').join('\n\t');
```
kyle-bowden, line 101: @JoPintoPaul 

Had a thought about how we could get rid of toString().split.join, we could probably treat the XML as an XML document then we don't need to worry about string functions.

Something like this

```
const pom = "Current pom XML";

// create XML document
let xmlDoc = parser.parseFromString(pom, "text/xml");

// example of how to set a node value
xmlDoc.getElementsByTagName('groupId')[0].textContent = "test"

// serialize XML document back to XML string
let s = new XMLSerializer();
let newXmlStr = s.serializeToString(xmlDoc);
```

tasks/gulp/webjar-helpers.js
```diff
@@ -120,7 +125,7 @@ const webJarHelpers = ({
     </scm>
 
     <dependencies>
```
oscarduignan, line 127: I'm curious to understand if this is going to cause projects to pull in the webjar for govuk-frontend and accessible-autocomplete that duplicates what we're packaging into the hmrc-frontend webjar, or if this is just descriptive. It seems dangerous if it pulls them in as dependencies because webjars put stuff onto the classpath so you could theoretically get some unexpected runtime stuff happening if there was overlap in paths between different webjars. I guess it won't cause a problem if there isn't a webjar version built - since we stopped publishing govuk-frontend webjar versions but the new versions since then don't seem to have caused a problem

tasks/gulp/webjar-helpers.js
JoPintoPaul: @kyle-bowden Agree, better to rewrite this using XML API. I propose creating a separate ticket to refactor this.

tasks/gulp/webjar-helpers.js
JoPintoPaul: @oscarduignan Agree, we should revisit this and see if the webjars are needed. Note we do publish the `govuk-frontend` webjar manually at the moment using npm webjars site. I propose creating a separate ticket to refactor this.

tasks/gulp/compile-js.js
```diff
@@ -56,3 +56,25 @@ gulp.task('js:compile-all-govuk-and-hmrc', () => gulp.src([
...
+
+gulp.task('js:compile-accessible-autocomplete', () => gulp.src([`${configPaths.src}accessible-autocomplete.js`])
+  .pipe(rollup({
+    plugins: [
+      babel(),
+      resolve(),
+      commonjs(),
+    ],
+  }, {
+    name: 'GOVUKAccessibleAutocomplete',
```
oscarduignan: With this, I wonder if we should make it match the govuk version, I'm thinking the prefix might confuse people using the documentation for accessible autocomplete, where it's ["accessibleAutocomplete"](https://github.com/alphagov/accessible-autocomplete/blob/main/webpack.config.babel.js#L115)

src/accessible-autocomplete.js
```diff
@@ -0,0 +1,3 @@
+import GOVUKAccessibleAutocomplete from 'accessible-autocomplete';
+
+window.GOVUKAccessibleAutocomplete = GOVUKAccessibleAutocomplete;
```
oscarduignan: @JoPintoPaul in the minified file, does it have a license header? wondering if we need to add this somewhere since we're redistributing https://github.com/alphagov/accessible-autocomplete/blob/main/LICENSE.txt wonder if we need

tasks/gulp/webjar-helpers.js
oscarduignan: yup separate ticket sounds right to me

tasks/gulp/compile-js.js
JoPintoPaul: @oscarduignan Interesting question. For me, the main reason to name it `GOVUKAccessibleAutocomplete` would be that it makes distinct our implementation of the JS / assets. 

At this time we happen to be using the `alphagov/accessible-autocomplete`, BUT if that library gets replaced, or we decide to swap in another implementation, our naming could end up being not aligned with our convention elsewhere but also not aligned with any other library.

On the other hand, it would be a simpler upgrade for teams to use the same library naming.

On balance though, my preference would be to name the import as `GOVUKAccessibleAutocomplete` as it's less coupled (if only by naming convention) to a specific implementation.

src/accessible-autocomplete.js
JoPintoPaul: @oscarduignan, hmm, you mean the GDS library uses the MIT licence and we use the Apache one?

tasks/gulp/compile-js.js
JoPintoPaul: @oscarduignan As discussed, this has been renamed `HMRCAccessibleAutocomplete` so that it is clear that our implementation may be different to the GDS implementation (at the moment it is the same, but we don't want to couple the naming).

src/accessible-autocomplete.js
JoPintoPaul: @oscarduignan As discussed - other uses of `govuk-frontend` aren't pulling in the MIT licence specifically, and in this case it is included via the `node_modules` folder. We'll discuss the GDS whether this is a requirement, and if so, we'll fix with the `govuk-frontend` as well in future work.

kyle-bowden (approved):

PR #250, opened 2022-09-21
PLATUI-1907 add back-link helper

ellamdav:


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [5.7.2] - 2022-09-21
```
JoPintoPaul: I think this should 5.8.0, not 5.7.2. I would say this is a minor version, not a hotfix.

docs/adr/0006-provide-back-link-helper.md
```diff
@@ -0,0 +1,42 @@
...
+
+* Teams have a canonical, tested solution that works by adding a single attribute
+* Teams don't need to copy/paste JS code into their service, or add inline scripts
+* The implementation can be centrally maintained, e.g. to cope with future browser API changes
+* Teams that are already using custom JS can, in time, remove it and use the new helper
+* The attack surface area of custom JS and inline scripts is reduced
+
+### Negative Consequences
+
+* Additional JS packaged and loaded for services that may not need it
```
JoPintoPaul, line 42:
```suggestion
* Additional JS packaged and loaded for services that may not need it
* By adding into the library, we are providing a tacit approval of using the Javascript back functionality, rather than encouraging teams to bind a bespoke URL within their service, which slightly negates the point of a back button
```

CHANGELOG.md
ellamdav: Agreed, that's just me on autopilot 😅

docs/adr/0006-provide-back-link-helper.md
ellamdav: Good point, well made.  Not sure if it's worth documenting that guidance somewhere, and maybe also signposting patterns like post-redirect-get?  I guess this is just because more and more teams are adding JS back links to their services, we might as well provide a packaged way to do it.

src/components/back-link-helper/back-link-helper.js
```diff
@@ -0,0 +1,30 @@
...
+BackLinkHelper.prototype.init = function init() {
+  // do nothing if History API is absent
+  if (this.window.history) {
+    // store referrer value to cater for IE
+    const docReferrer = this.document.referrer;
+
+    // prevent resubmit warning
+    if (this.window.history.replaceState && typeof this.window.history.replaceState === 'function') {
+      this.window.history.replaceState(null, null, this.window.location.href);
+    }
```
oscarduignan, line 16: I wonder if we'd want this as a separate thing, I could imagine wanting this on all pages, even where back links are not mimicking the browser back button via javascript

oscarduignan: (future work)

oscarduignan (approved):

PR #253, opened 2022-09-29
PLATUI-1830: Productionize the accessible autocomplete component in play-frontend-hmrc

kyle-bowden:


src/components/accessible-autocomplete/README.md
```diff
@@ -0,0 +1,64 @@
+# Accessible Autocomplete
+
+This macro takes an `<select>` and uses progressive enhancement, provided by the govuk accessible-autocomplete enhanceSelectElement function.
```
ellamdav: typo
```suggestion
This macro takes a `<select>` and uses progressive enhancement, provided by the govuk accessible-autocomplete enhanceSelectElement function.
```

src/components/accessible-autocomplete/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Look [here](https://www.npmjs.com/package/accessible-autocomplete) for more information on the attributes for the accessible autocomplete component.
+
+## id
+Should be the same as the id for this component
+
+## data-default-value
+Type: Boolean
+
+If this is set to true, all values are shown when the user clicks the input. This is similar to a default dropdown, so the autocomplete is rendered with a dropdown arrow to convey this behaviour
```
ellamdav: This looks like the type/description for `data-show-all-values` - is this right?

kyle-bowden: Thanks for spotting that, i have updated it to the correct description.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [5.11.0] - 2022-09-29
+
+### Added
+
+- Add new [accessible-autocomplete](src/components/accessible-autocomplete) component to provide a common way to use the govuk accessible-autocomplete component
```
```suggestion
- Added way to automatically enhance a select element into an [accessible-autocomplete](src/components/accessible-autocomplete) using the `data-module="hmrc-accessible-autocomplete"` data attribute, more details in the [accessible-autocomplete readme](src/components/accessible-autocomplete/README.md)
```

src/components/accessible-autocomplete/accessible-autocomplete.yaml
```diff
@@ -0,0 +1,116 @@
...
+  - name: hint
+    type: object
+    required: false
+    description: Options for the hint component.
+  - name: describedBy
+    type: string
+    required: false
+    description: One or more element IDs to add to the `aria-describedby` attribute, used to provide additional descriptive information for screenreader users.
+
+examples:
```
oscarduignan, line 23: @kyle-bowden just had a thought, could we have an example with just the data-module attribute, so none of the other data attribute options provided? We have examples of different combinations of attributes but not one where the optional ones are omitted all together

kyle-bowden: I have added an additional example that takes a select with no optional data attributes, just the `data-module="hmrc-accessible-autocomplete"` attribute. For the integration test, I test that the enhanceSelectElement generates an input with all default values, as I believe that should be the behavior.

oscarduignan (approved):

PR #254, opened 2022-10-06
NOJIRA: Stop accessible-autocomplete css including webfonts

oscarduignan:
where we [import the helpers](https://github.com/hmrc/hmrc-frontend/blob/main/src/accessible-autocomplete.scss#L3) it's pulling in the [fonts](https://github.com/alphagov/govuk-frontend/blob/4b5ed0ada3b64722ae2074dfa0db5d4d3a45cb79/src/govuk/helpers/_font-faces.scss) - we need to set `$govuk-include-default-font-face: false;` ([which stops it from being pulled in when we use govuk-font()](https://github.com/alphagov/govuk-frontend/blob/e319f83cc2496e6814e0eb8f341ae658b05b415d/package/govuk/helpers/_typography.scss#L26-L28)) before we do the `@import` in that file to stop it. Because the font will already be available on the page via the main hmrc-frontend css

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [5.11.1] - 2022-10-06
+
+### Fixed
+
+- Stopped the accessible-autocomplete css from including the GDS Transport webfonts because we assume they will already be loaded
+
 ## [5.11.0] - 2022-09-29
 
 ### Added
 
 - Added way to automatically enhance a select element into an [accessible-autocomplete](src/components/accessible-autocomplete) using the `data-module="hmrc-accessible-autocomplete"` data attribute, more details in the [accessible-autocomplete readme](src/components/accessible-autocomplete/README.md)
-
 
 ## [5.10.0] - 2022-09-26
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,13 +7,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [5.12.0] - 2022-10-06
```
ellamdav:
```suggestion
## [5.11.1] - 2022-10-06
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "5.11.0",
+  "version": "5.12.0",
```
ellamdav:
```suggestion
  "version": "5.11.1",
```

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "5.11.0",
+  "version": "5.12.0",
```
ellamdav:
```suggestion
  "version": "5.11.1",
```

ellamdav (commented):
LGTM but should probably be a bugfix release?

ellamdav (approved):

PR #258, opened 2022-11-23
PLATUI-2101: fix our css rules for autocomplete when autoselect option is used

KatiePocklington:


src/accessible-autocomplete.scss
```diff
@@ -6,6 +6,7 @@ $govuk-include-default-font-face: false;
 @import "../../accessible-autocomplete/dist/accessible-autocomplete.min";
 
 .autocomplete__wrapper,
-.autocomplete__input {
+.autocomplete__input,
+.autocomplete__hint {
```
oscarduignan, line 10: @KatiePocklington / @kyle-bowden could you create a visual regression test for this?

oscarduignan: some details about how you can do it in the contributing doc https://github.com/hmrc/hmrc-frontend/blob/main/CONTRIBUTING.md#visual-regression-tests

oscarduignan (approved):

PR #262, opened 2022-11-29
PLATUI-2106: fix timeout dialog bug

oscarduignan:
Had to update puppeteer and tweak account-menu test as a result

I'm off the next 3 days but 🤞 these give a good test harness for fixing the bug

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "5.13.0",
+  "version": "5.14.0",
   "description": "Design patterns for HMRC frontends",
   "scripts": {
     "start": "gulp dev",
@@ -48,6 +48,7 @@
   },
   "devDependencies": {
     "@babel/preset-env": "^7.18.10",
+    "@sinonjs/fake-timers": "^10.0.0",
     "adm-zip": "^0.5.5",
     "adr-log": "^2.2.0",
     "ajv": "^6.12.2",
@@ -55,6 +56,7 @@
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
     "backstopjs": "^6.0.4",
     "better-npm-audit": "^3.7.3",
+    "browserify": "^17.0.0",
     "chalk": "^4.1.0",
     "cheerio": "^1.0.0-rc.10",
     "cssnano": "^4.1.10",
@@ -89,7 +91,7 @@
     "oldie": "^1.3.0",
     "postcss-pseudo-classes": "^0.2.0",
     "postcss-scss": "^2.0.0",
-    "puppeteer": "^1.19.0",
+    "puppeteer": "19.3.0",
     "recursive-readdir": "^2.2.2",
     "remarkable": "1.7.4",
     "request": "^2.88.2",
--- a/src/components/timeout-dialog/timeout-dialog.js
+++ b/src/components/timeout-dialog/timeout-dialog.js
@@ -17,12 +17,6 @@ function TimeoutDialog($module, $sessionActivityService) {
   let currentTimer;
   const sessionActivityService = $sessionActivityService;
 
-  cleanupFunctions.push(() => {
-    if (currentTimer) {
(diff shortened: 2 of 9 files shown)


src/components/timeout-dialog/__tests__/timeout-dialog-can-be-shown-multiple-times.test.js
```diff
@@ -0,0 +1,148 @@
...
+    if (req.url() === url) {
+      req.respond({
+        contentType: 'text/html',
+        body,
+      });
+    } else {
+      req.continue();
+    }
+  });
+}
```
oscarduignan: not used anywhere else at the moment but could move these into `__tests__/__helpers__/*`

oscarduignan: thinking about it some more, might be better to rename `__helpers__` to `__test-helpers__` and move this test file alongside the other test files (up one level, out of `__tests__`). Could refactor our tests into `__tests__` folder in future, but at the moment we have convention of having them inline

src/components/timeout-dialog/__tests__/__helpers__/install-fake-timers.js
```diff
@@ -0,0 +1,3 @@
+const FakeTimers = require('@sinonjs/fake-timers');
```
oscarduignan: need to add a comment to explain how this is being used

package.json
```diff
@@ -89,7 +91,7 @@
     "oldie": "^1.3.0",
     "postcss-pseudo-classes": "^0.2.0",
     "postcss-scss": "^2.0.0",
-    "puppeteer": "^1.19.0",
+    "puppeteer": "19.3.0",
```
oscarduignan, line 94: can't quite believe how far behind we were, guess maybe puppeteer changed their version numbering

oscarduignan:
tests failed in jenkins but pass locally, think maybe we need to wait for the element to be visible in the tests, a tomorrow problem

oscarduignan:
tests failing on jenkins wasn't a race condition, it was because of hardcoded version number in the test fixtures

src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -17,12 +17,6 @@ function TimeoutDialog($module, $sessionActivityService) {
   let currentTimer;
   const sessionActivityService = $sessionActivityService;
 
-  cleanupFunctions.push(() => {
-    if (currentTimer) {
-      window.clearTimeout(currentTimer);
-    }
-  });
```
oscarduignan, line 24: each time the cleanup function runs it mutates the cleanupFunctions array and one by one pulls items from it and executes them until the array is empty

when this cleanup function is added here, it means that when it is cleaned up once it will never be cleaned up again, and the next time the currentTimer was created it would run to completion even if you hit "keep signed in"

the currentTimer is the timer that's created when the dialog is opened, and redirects the user to the timeout url, after they timeout, if the user doesn't interact with the timeout warning

ellamdav (approved):

src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -143,6 +137,9 @@ function TimeoutDialog($module, $sessionActivityService) {
 
     cleanupFunctions.push(() => {
       window.clearTimeout(timeout);
+      if (currentTimer) {
+        window.clearTimeout(currentTimer);
+      }
```
oscarduignan, line 142: moving the cleanup here means that a cleanup function to clear the current timer is created each time the dialog is opened. currentTimer is created in the recursive function called a the end of setupDialog

oscarduignan:
have the acceptance tests but realised I haven't really explained why moving the bit of code that we did fixes the problem, so I've added some comments to the PR. The whole timeout dialog code could probably use some refactoring, but I'm not confident that we won't introduce bugs if we do any more now so kept it lightest touch possible

PR #276, opened 2023-02-08
PLATUI-2116: Updated header to use localised text for menu in text an…

JoPintoPaul:
…d aria-label


oscarduignan (dismissed):

src/components/header/header.yaml
```diff
@@ -158,6 +158,21 @@ examples:
       - href: '#4'
         text: Navigation item 4
 
+- name: with navigation welsh
```
timsb, line 161: Not sure if this affects anything but is it worth having a Welsh example with a longish service name? Just thinking because of the extra characters in the word 'Dewislen' we might want to make sure there's no overlapping of text. I think it's not a problem at the moment because things are floated (If I remember correctly) but if it changes in the future we might not spot it if there's no VRTs.

timsb (approved):

oscarduignan (approved):

PR #288, opened 2023-04-17
PLATUI-2282

oscarduignan:
When timeout is reached, it will keep counting down past 0 and if the
timeout page takes longer than the 1-second tick then the timeout
redirection gets cancelled. Instead when the timeout is reached we want
to stop doing anything and let the timeout redirection happen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [5.28.0] - 2023-04-20
+
+### Fixed
+
+- Updated timeout-dialog to fix bug. If user timed out and the timeout redirection took longer than a second then it
+  would get canceled and user would stay on page while the timeout dialog counted down into negative numbers.
+
 ## [5.27.0] - 2023-03-20
 
 ### Changed
--- a/src/components/timeout-dialog/timeout-dialog-can-be-shown-multiple-times.test.js
+++ b/src/components/timeout-dialog/timeout-dialog-can-be-shown-multiple-times.test.js
@@ -10,6 +10,49 @@ import { installFakeTimersOnLoad, serveFakePage } from './test-helpers/browser-t
 import { version } from '../../../package.json';
 
 describe('timeout dialog with a single page open', () => {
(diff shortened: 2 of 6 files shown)


src/components/timeout-dialog/timeout-dialog-can-be-shown-multiple-times.test.js
oscarduignan, line 1: the name for this file is probably too narrow with the addition of this extra test, but I think that refactoring the browser based tests in this repo should go into tech debt board and be done in bulk because the setup is a bit creaky. We should refactor to use jest-puppeteer and probably fold some separate test files into one another / decide on a different file structure for separating tests

src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -266,12 +266,13 @@ function TimeoutDialog($module, $sessionActivityService) {
     };
 
     const runUpdate = () => {
-      const counter = getSecondsRemaining();
+      const counter = Math.max(getSecondsRemaining(), 0);
       updateCountdown(counter);
-      if (counter <= 0) {
+      if (counter === 0) {
         timeout();
+      } else {
```
oscarduignan, line 273: the `else` is the important bit, the bounding to 0 is because once there are no seconds remaining in the session it only makes sense to say that, but we theoretically (and because of this bug) could show negative seconds remaining

oscarduignan: in terms of cross browser compat, Math.max is supported everywhere https://caniuse.com/mdn-javascript_builtins_math_max

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [5.27.0] - 2023-04-17
```
kyle-bowden: Is this not meant to be version 5.28.0

oscarduignan: totally forgot to do this! 🤦

oscarduignan: DONE!

kyle-bowden (approved):

PR #292, opened 2023-05-02
PLATUI-2304: Add empty alt attribute to logo IE8 fallback PNG in head…

JoPintoPaul:
…er, aligning with change in govuk-frontend

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "5.30.0",
+  "version": "5.31.0",
   "description": "Design patterns for HMRC frontends",
   "scripts": {
     "start": "gulp dev",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "5.30.0",
+  "version": "5.31.0",
```
timsb, line 3: Wondering if this could be classed as a patch i.e. 5.30.1?
Looking at https://semver.org/ they say 'Bug fixes not affecting the API increment the patch version, backwards compatible API additions/changes increment the minor version, and backwards incompatible API changes increment the major version.'. I think this could be classed as an accessiblity bug fix as it doesn't affect the API in any way?

timsb (changes requested):

timsb (approved):

PR #294, opened 2023-05-10
PLATUI-2305: Removed references to check-compatability

JoPintoPaul:


src/govuk-prototype-kit.config.test.js
```diff
@@ -12,6 +12,6 @@ const configuredComponents = require('./govuk-prototype-kit.config.json')
 
 describe('govuk-prototype-kit config', () => {
   it('should configure all HMRC components for use with v13+ GOVUK prototype kit', () => {
-    expect(configuredComponents).toEqual(definedComponents);
+    expect(configuredComponents).toEqual(definedComponents.reverse());
```
timsb, line 15: How come this now has to be reversed? Is it something to do with one of our dependencies?

JoPintoPaul: @timsb It is, the order is now the other way around. As this is an array I assumed it wouldn't cause problems, but if it would (or of there's a better way to test), happy to pair on this.

timsb (approved):

src/govuk-prototype-kit.config.test.js
timsb: I think it's readable. Was just interested as to why that was needed now!

PR #296, opened 2023-05-25
PLATUI-2336 Increase test coverage before enabling new gov.uk link st…

KatiePocklington:
…yles


src/components/account-header/account-header.yaml
```diff
@@ -87,6 +87,7 @@ examples:
       classes: 'the-header-classes'
   visualRegressionTests:
     backstopScenarioOptions:
+      misMatchThreshold: 0.1
```
oscarduignan, line 90: for some reason, even though tests are being done in docker container, aand katie is on linux like jenkins, in this test, the account menu hover on language switcher has a 0.03 change in pixels (underline is lower in one than other) https://build.tax.service.gov.uk/job/PlatUI/job/Libraries/job/hmrc-frontend-pr-builder/193/artifact/backstop_data/html_report/index.html and search for:
```
account-header (link-states-on-desktop example) (language-select-current-lang-focused-and-alternate-lang-hovered)
```
guessing we might not need this anymore once we enable the new link styles, or this could be something we could fix by reviewing our custom link styles for this and figuring out what is causing this difference in rendering

kyle-bowden, line 90: I'm guessing this was introduced as there is probably a pixel that is out of place when hovering occurs.

kyle-bowden: So are happy to get this change in then? looks good to me.

kyle-bowden (approved):

backstop_data/bitmaps_reference/backstop_default_account-menu_with-navigation-urls_example_0__0_phone.png
oscarduignan, line 1: all these screenshots were old and not being used, it seems like the name of the reference data changed - probably as I was changing the way that VRT tests were configured and their labels https://github.com/hmrc/hmrc-frontend/blob/fab0ec495ecd39e0ab2d14e4170396c3338b088c/tasks/gulp/backstop-config.js#L47 and I probably manually updated the reference folder or something like that - so they weren't hurting anything - but they were just random photos of lemur butts living in our bitmaps_reference folder for no reason

src/components/account-header/account-header.yaml
oscarduignan: yup, it's it's just the underline moving a pixel down - guess maybe some number rounding thing or something like that - no issues for anyone using the component

PR #310, opened 2023-07-28
PLATUI-2411 always show Account Menu nav buttons if JS is disabled

ellamdav:


src/components/account-menu/_account-menu.scss
```diff
@@ -13,6 +13,7 @@
 // Necessary as there's a specicifity problem on the <li>'s in the mobile menu
 .hidden,
 .no-js .js-visible,
+.js-enabled .js-hidden,
```
ellamdav, line 16: Added this in place of the broader `js-hidden` as we only want to hide these if JS is enabled.  The existing `.js` and `.no-js` look to be vestigial things, or at least I can't see where they are being used, but I wasn't comfortable just removing these for fear of breaking some service that relies on this.

oscarduignan (approved):

PR #311, opened 2023-08-01
Add pluginDependencies to govuk-prototype-kit.config.json

oscarduignan:
Prompted by @nataliecarey 🙌 this helps ensure that people installing hmrc-frontend via the govuk-prototype-kit management interface also get govuk-frontend

Turning off the git commit & tag behaviour of `npm version minor` doesn't impact our release process because we tag releases in a different way.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,8 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npm run lint
+
+node lib/update-govuk-prototype-kit-config.js
+
+git add src/govuk-prototype-kit.config.json
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
+git-tag-version=false
--- a/lib/update-govuk-prototype-kit-config.js
+++ b/lib/update-govuk-prototype-kit-config.js
@@ -1,0 +1,14 @@
+const path = require('path');
+const fs = require('fs');
+
+const projectDir = path.dirname(__dirname);
+const govukFrontendVersion = JSON.parse(fs.readFileSync(path.join(projectDir, 'package.json'), 'utf8')).dependencies['govuk-frontend'].replace(/[^0-9.]/g, '');
+const govukPrototypeKitConfigPath = path.join(projectDir, 'src', 'govuk-prototype-kit.config.json');
+const govukPrototypeKitConfig = JSON.parse(fs.readFileSync(govukPrototypeKitConfigPath, 'utf8'));
+govukPrototypeKitConfig.pluginDependencies = [
+  {
+    packageName: 'govuk-frontend',
+    minVersion: govukFrontendVersion,
+  },
+];
+fs.writeFileSync(govukPrototypeKitConfigPath, JSON.stringify(govukPrototypeKitConfig, null, 2));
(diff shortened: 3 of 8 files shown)


JoPintoPaul (dismissed):

lib/update-govuk-prototype-kit-config.js
oscarduignan, line 1: although it's kind of a build task, at the moment our build tasks are all still gulp - so rather than have a non-gulp build task hiding in `tasks/` I opted to put it in lib which is also [not included in package](https://github.com/hmrc/hmrc-frontend/blob/main/tasks/gulp/package.js)

src/govuk-prototype-kit.config.json
```diff
@@ -1,4 +1,10 @@
 {
+  "pluginDependencies": [
+    {
+      "packageName": "govuk-frontend",
+      "minVersion": "4.7.0"
+    }
+  ],
```
oscarduignan, line 7: when you install hmrc-frontend through the prototype kit UI this will prompt the designer that govuk-frontend is also needed (if it isn't already installed) and it can then install it for them. Otherwise if govuk-frontend wasn't installed someone could install hmrc-frontend without realising they also needed govuk-frontend and not get stuff autowired together with the prototype kit conventions. It's not a likely scenario at the moment, but here's an example of what the prompt looks like when installing via the web ui of the prototype kit (though this example is installing the govuk frontend common templates plugin not hmrc-frontend, but it would look the same):
CleanShot 2023-08-01 at 11 36 02@2x

.npmrc
```diff
@@ -0,0 +1 @@
+git-tag-version=false
```
oscarduignan, line 2: Turning off the git commit & tag behaviour of `npm version minor` doesn't impact our release process because we tag releases in a different way.

.husky/pre-commit
oscarduignan, line 1: our husky hooks got turned off in the past, or updated to a new version where they didn't work - I've turned them on again but only the pre-commit because the pre-push one that ran the tests would prevent some contributors pushing if they couldn't replicate the visual regression test results

originally I had the update-govuk-prototype-kit-config in `npm version` script but realised as I was about to share this with Tim that it would work in a husky pre-commit hook too and then never be able to get out of date

package.json
```diff
@@ -123,11 +125,5 @@
     "testRunner": "jest-jasmine2",
     "globalSetup": "./lib/puppeteer/setup.js",
     "globalTeardown": "./lib/puppeteer/teardown.js"
-  },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm run lint",
-      "pre-push": "npm test"
-    }
```
oscarduignan, line 131: these weren't being run, I think husky had been removed - possibly because the pre-push hook was blocking contributions from m2 macs where the results from VRT were different

.npmrc
```diff
@@ -0,0 +1,2 @@
+PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
```
oscarduignan, line 1: we don't use playwright, but backstop has a dependency on it and when they added that all our build started downloading binaries for all playwright browsers unnecessarily - this disable that behaviour

.husky/pre-commit
oscarduignan, line 1: originally had this in an `npm version` package json script, but realised that it could run in a pre-commit hook safely - that lead to discovery that husky was removed so I re-enabled it partially (just pre-commit hook)

timsb (dismissed):

JoPintoPaul (approved):

ellamdav (approved):

PR #314, opened 2023-08-17
PLATUI-2381: test using selenium for VRT

oscarduignan:
not runnable on CI yet, but probably can be - though will require some build jobs changes - as it seems like you can't specify other ports to map at the moment but this switches backstop to use playwright and makes playwright use selenium to see if using selenium-node docker containers on build and local will result in reproducible VRT tests for us (and handily make our builds more idiomatic to platform guidelines of using selenium)

test with:
```
nvm install
npm install
./run-docker-browser.sh
npm run backstop:test
```

bits I'm unsure about:
- can arm architectures like new mac run the browser containers?
- if they can run the browser containers, will the backstop results be reproducible?
- will browser results locally be reproducible with results in build sidecars?


KatiePocklington:
> npm run backstop:test

PR #315, opened 2023-08-17
Added plugin metadata

nataliecarey:
We're working on a new plugin page, we'll be making use of new metadata that comes from plugin config.  This PR adds that metadata.

I've also added the validator to the post-build tests, this uses `npx` and `@latest` to always look at the latest requirements.  I recommend this approach but it does erode the purity of repeatable builds, you could hardcode a version number in there or install `govuk-prototype-kit` as a (dev) dependency and remove both `npx` and `@latest`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Design patterns for HMRC frontends",
   "scripts": {
     "start": "gulp dev",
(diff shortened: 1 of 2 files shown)


hmrc-web-operations:

oscarduignan (approved):
Thank you @nataliecarey 🙌 I reckon having @latest is probably pretty safe since we never really rebuild or patch old versions

PR #318, opened 2023-09-15
PLATUI-2567 revert general js-hidden change that was impacting teams; replace with CSS specific to the account menu

ellamdav:


src/components/account-menu/_account-menu.scss
```diff
@@ -349,4 +355,9 @@ $hmrc-banner-blue: #005ea5;
   }
 }
 
+// Note: this is relied on by teams not using the account menu, so handle with care
+.js-hidden {
+  display: none;
+}
```
oscarduignan, line 361: might be odd but relied on behaviour, but this will always hide the .js-hidden elements even if there is no js

oscarduignan: oh I get it, it's not saying "this will be hidden when there is js enabled" it's saying "this is a class we add to stuff with js to hide it"

oscarduignan: https://github.com/hmrc/assets-frontend/blob/f0b2b9830405199345e0f7b5abd710348b14cc8c/assets/javascripts/modules/mask.js#L108-L109

oscarduignan: but would have thought that 
```
.js-enabled .js-hidden
```
would cover it in that case - are people relying on js-hidden to hide stuff when js is disabled? that's the only case I can think this wouldn't work - or that people aren't using the govuk layout and so not getting js-enabled added

oscarduignan (dismissed):

kyle-bowden (approved):

PR #320, opened 2023-09-27
PLATUI-2554 Add utility class to hide stuff when javascript is disabled 

KatiePocklington:


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [5.49.0] - 2023-09-27
+
+### Changed
+
+- Add utility class to stuff when javascript is disabled
```
```suggestion
- Add utility class to hide stuff when javascript is disabled
```

tmc-ee (dismissed):

src/components/js-visible/README.md
```diff
@@ -0,0 +1,3 @@
+# Utility class to hide stuff when javascript is disabled 
+
+It assumes that you are using the GOV.UK template that adds the js-enabled class to the body element
```
oscarduignan: Just had a thought of some extra context we can add to this

```suggestion
It assumes that you are using the GOV.UK template that adds the js-enabled class to the body element 

If the js-enabled class is not added to the body element of the page with javascript then the element will always be hidden

The elements are hidden using `display: none;` so will not be read aloud by screen readers or navigable by keyboard when hidden
```

oscarduignan:
we decided it wasn't worth having an automated test for this because it's a utility class and the bit where it integrates with the layout by depending on js-enabled is something that won't be removed or changed in govuk template casually, because it would be a big breaking change that would be shared loudly

oscarduignan (approved):

PR #324, opened 2023-10-31
PLATUI-2588: Adding in hideSignOutLink functionality

JoPintoPaul:


src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -182,20 +185,24 @@ function TimeoutDialog($module, $sessionActivityService) {
...
     $element.appendChild(document.createTextNode(' '));
-    $element.appendChild(wrapLink($signOutButton));
+
+    if (!settings.hideSignOutLink) {
+      const $signOutButton = utils.generateDomElementFromStringAndAppendText(
+        '<a id="hmrc-timeout-sign-out-link" class="govuk-link hmrc-timeout-dialog__link">',
+        settings.signOutButtonText,
+      );
+
+      $staySignedInButton.addEventListener('click', keepAliveAndClose);
```
oscarduignan: the stay signed in button still needs it's event listener added I think

src/components/timeout-dialog/timeout-dialog.test.js
```diff
@@ -336,6 +336,35 @@ describe('/components/timeout-dialog', () => {
...
+        'data-keep-alive-button-text': 'KEEP alive',
+        'data-hide-sign-out-link': true,
+      });
+
+      pretendSecondsHavePassed(780);
+
+      expect(testScope.latestDialog$element.querySelector('a#hmrc-timeout-sign-out-link')).toBeNull();
+    });
+
+    it('should show sign out button when disable-sign-out-button parameter is unset', () => {
```
ellamdav: maybe make it clear that this is the default?
```suggestion
    it('should show sign out button by default', () => {
```

src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -182,20 +185,24 @@ function TimeoutDialog($module, $sessionActivityService) {
...
 
     $element.appendChild($visualMessge);
     $element.appendChild($audibleMessage);
     $element.appendChild($staySignedInButton);
     $element.appendChild(document.createTextNode(' '));
-    $element.appendChild(wrapLink($signOutButton));
+
+    if (!settings.hideSignOutLink) {
+      const $signOutButton = utils.generateDomElementFromStringAndAppendText(
+        '<a id="hmrc-timeout-sign-out-link" class="govuk-link hmrc-timeout-dialog__link">',
```
ellamdav, line 197: we seem to have a mix of (existing) names - `signOutButton` vs `sign-out-link` which is perhaps confusing for the end user, but maybe not worth fixing now.

ellamdav (dismissed):

src/components/timeout-dialog/timeout-dialog.yaml
```diff
@@ -48,6 +48,7 @@ examples:
     keepAliveButtonText: "keepAliveButtonText"
     signOutButtonText: "signOutButtonText"
     synchroniseTabs: false
+    hideSignOutLink: false
```
oscarduignan: we could add a comment here maybe, `this is experimental - don't use without testing, added at request of non-hmrc service, where a service only has an anonymous session, so no explicit log in action has been taken, and you might want to have no early sign out link rather along with the than a link saying something like [abandon attempt]`

src/components/timeout-dialog/timeout-dialog.yaml
```diff
@@ -48,6 +48,10 @@ examples:
     keepAliveButtonText: "keepAliveButtonText"
     signOutButtonText: "signOutButtonText"
     synchroniseTabs: false
+    hideSignOutButton: false
+#   hideSignOutButton is experimental, please don't use in HMRC. without testing. Added at request of non-HMRC service,
+#   where a service only has an anonymous session, so no explicit log in action has been taken, and you might want to
+#   have no early sign out link rather along with the than a link saying something like [abandon attempt]
```
oscarduignan: sorry, reread and my suggestion before was bit verbose 😅 what about this
```suggestion
    hideSignOutButton: false
#   hideSignOutButton is experimental, please don't use in HMRC services without testing with your users, see github issues/316 
```

oscarduignan (approved):

PR #325, opened 2023-11-01
PLATUI-2590: Make a link or button print the page without having to write any custom javascript

kyle-bowden:


src/components/hmrc-print-link/example.njk
```diff
@@ -0,0 +1,16 @@
...
+<a href="#" data-module="hmrc-print-link">Print this page</a>
+</br>
+</br>
+<h3>Input Button</h3>
+<input type="button" data-module="hmrc-print-link" value="Print this page"/>
+</br>
+</br>
+<h3>Button</h3>
+<button type="button" data-module="hmrc-print-link">Print this page</button>
+
```
oscarduignan, line 6: small one, I think (mostly in case someone copy pastes it) that on these we should add the govuk-link and govuk-button classes - without any font styles the VRT screenshots might not be exactly the same too I think when captured on jenkins

oscarduignan: a small extra bit - it might be worth one of these examples using the govukButton component so show how you add the attribute via it's params

src/all.js
```diff
@@ -4,13 +4,29 @@ import TimeoutDialog from './components/timeout-dialog/timeout-dialog';
...
+  const $HmrcPrintLinkButtonSelector = 'button[data-module="hmrc-print-link"]';
+  if (document.querySelector($HmrcPrintLinkButtonSelector)) {
+    new HmrcPrintLink($HmrcPrintLinkButtonSelector, window).init();
+  }
+
+  const $HmrcPrintLinkInputButtonSelector = 'input[type="button"][data-module="hmrc-print-link"]';
+  if (document.querySelector($HmrcPrintLinkInputButtonSelector)) {
+    new HmrcPrintLink($HmrcPrintLinkInputButtonSelector, window).init();
+  }
+
```
oscarduignan, line 19: it's unlikely that people will use more than one of these on a page, but maybe we could loop through these instead like we do for back links. You can also provide a comma separated list of selectors to querySelectorAll so rather than these three. The convention of the other ones too is that rather than the module doing the initial query selecting, you do it here, so `const printLinks = document.querySelectorAll("...")`
```
  const $BackLinks = document.querySelectorAll('[data-module="hmrc-back-link"]');
  nodeListForEach($BackLinks, ($BackLink) => {
    new BackLinkHelper($BackLink, window, document).init();
  });
```

oscarduignan:
in the component folder we could add a README.md that quickly explained why we have this component - I know that people can see the jira / github commit message but since jira's days are possibly numbered 🤷‍♂️  maybe something like
> we've added this because a lot of services had implemented print links using inline javascript in html attributes that isn't compatible with some new default CSP security config that is recommended for HMRC services. Rather that everyone move the JS to different places / possibly introduce mistakes - this gives us something people can migrate to a bit more easily, and we now have people depending on an interface we can tweak later in one place in the remote chance we need

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [5.54.0] - 2023-11-02
+
+### Changed
+
+- Provided a data attribute to allow links and buttons to call window.print
```
```suggestion
- Provided a data attribute to allow links to call window.print without services having to write any custom javascript
```

src/all.js
```diff
@@ -4,13 +4,19 @@ import TimeoutDialog from './components/timeout-dialog/timeout-dialog';
...
 function initAll() {
   const $AccountMenuSelector = '[data-module="hmrc-account-menu"]';
   if (document.querySelector($AccountMenuSelector)) {
     new AccountMenu($AccountMenuSelector).init();
   }
 
+  const $HmrcPrintLink = document.querySelector('a[data-module="hmrc-print-link"]');
+  if ($HmrcPrintLink) {
+    new HmrcPrintLink($HmrcPrintLink, window).init();
+  }
```
oscarduignan:
```suggestion
  const $HmrcPrintLinks = document.querySelectorAll('a[data-module="hmrc-print-link"]');
  nodeListForEach($HmrcPrintLinks, ($HmrcPrintLink) => {
    new HmrcPrintLink($HmrcPrintLink, window).init();
  });
```

just in case for some reason someone puts two on one page, like at top and bottom of a long page maybe

src/components/hmrc-print-link/hmrc-print-link.yaml
```diff
@@ -0,0 +1,19 @@
...
+    required: false
+    description: id to use within the govuk button component.
+  - name: text
+    type: string
+    required: false
+    description: Text to use within the govuk button component.
+  - name: attributes
+    type: object
+    required: false
+    description: HTML attributes (for example data attributes) to add to the anchor tag.
```
oscarduignan: sorry I didn't spot this before 😮‍💨 I'm going to approve anyway since I think no-one ever looks at params and probably wouldn't expect to for this, but if you do feel like it:

```suggestion
params: []
```

oscarduignan (dismissed):

oscarduignan (approved):

PR #329, opened 2023-11-06
PLATUI-2638 Enable Welsh language support for accessible-autocomplete

ellamdav:
* Pass Welsh language functions to accessible-autocomplete if `data-language` is set to `cy`
* Updated README for the component to reflect current API and strip out irrelevant content


TimothyFothergill (dismissed):
lgtm

oscarduignan:
looks good to me, I guess the alternative would be the GDS convention of having translations provided to js stuff via data-i18n prefixed attributes `data-i18n.assistive-hint="..."` https://github.com/alphagov/govuk-frontend/blob/4d13a759e4ffdea97309fde322eeb82087ac6e23/packages/govuk-frontend/src/govuk/macros/i18n.njk#L17-L21 but then we wouldn't have any welsh translations out of the box in hmrc-frontend, they would have to be in play-frontend-hmrc

oscarduignan (approved):

PR #331, opened 2023-11-20
PLATUI-2644: Hide when printing for language select and report techni…

JoPintoPaul:
…cal issue

### DIFF
--- a/src/components/language-select/template.njk
+++ b/src/components/language-select/template.njk
@@ -1,5 +1,5 @@
 {% if params.language === 'en' %}
-  <nav class="hmrc-language-select" aria-label="Language switcher">
+  <nav class="hmrc-language-select govuk-!-display-none-print" aria-label="Language switcher">
     <ul class="hmrc-language-select__list">
       <li class="hmrc-language-select__list-item">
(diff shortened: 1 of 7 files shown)


TimothyFothergill (dismissed):
lgtm

src/components/language-select/template.njk
```diff
@@ -1,5 +1,5 @@
 {% if params.language === 'en' %}
-  <nav class="hmrc-language-select" aria-label="Language switcher">
+  <nav class="hmrc-language-select govuk-!-display-none-print" aria-label="Language switcher">
```
ellamdav, line 2: worth adding tests for these?

JoPintoPaul: @ellamdav Could do, could do... Seems like it's just checking the static template rather than any logic, but can add if that's what we think is best?

ellamdav: It feels like double-entry coding in a way, but that's not such a bad thing - eg. someone could refactor the template to dry it up and this might get lost, but visually would still all look OK 🤷

ellamdav (approved):

PR #336, opened 2024-01-03
PLATUI-2719: Added new hmrc-frontend-print-overrides stylesheet, remo…

JoPintoPaul:
…ved hardcoded govuk-display-none-print, added ADR

### DIFF
--- a/docs/adr/0007-add-print-stylesheet.md
+++ b/docs/adr/0007-add-print-stylesheet.md
@@ -1,0 +1,49 @@
+# Add print stylesheet
+
+* Status: accepted
+* Deciders: PlatUI
+* Date: 2024-01-03
+* Components: [hmrc-frontend-print-overrides.scss](/src/hmrc-frontend-print-overrides.scss)
+
+## Context and Problem Statement
+
+In August 2022, CSG Forms (formerly known as Print & Post) opened a support issue with PlatUI asking for various parts
+of page components to be hidden from print views. This had been flagged as an issue by an internal accessibility audit.
+
+`govuk-frontend` provides a helper class for hiding elements from print views (`govuk-!-display-none-print`). This was 
+applied to the Nunjucks for some `hmrc-frontend` components over time. However, there were technical complexities around
+adding the same class to `govuk-frontend` components without breaking library parity between `govuk-frontend` and 
+`play-fronted-hmrc`. 
+
+## Decision Drivers
+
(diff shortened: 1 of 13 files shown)


src/hmrc-frontend-print-overrides.scss
```diff
@@ -0,0 +1,13 @@
...
+.govuk-back-link,
+.govuk-summary-list__actions,
+.govuk-footer__section,
+.govuk-footer__inline-list,
+.hmrc-report-technical-issue,
+.hmrc-language-select {
+  @include govuk-media-query($media-type: print) {
+    display: none !important;
+  }
+}
```
oscarduignan, line 10: I think it might be worth removing the imports and just having a plain
```
@media print {
  .govuk-back-link,
  .govuk-summary-list__actions,
  .govuk-footer__section,
  .govuk-footer__inline-list,
  .hmrc-report-technical-issue,
  .hmrc-language-select {
    display: none !important;
  }
}
```

oscarduignan: (context for why they have the govuk mixin - I think - is that it lets you target a breakpoint by name - so you can say "desktop" and it has a k/v map that translates that into the viewport dimensions you need to apply styles at to cover that alias - but if we're just targeting by media then it's not doing more than a css `@media {}`)

docs/adr/0007-add-print-stylesheet.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Positive Consequences
+
+* It provides a centrally-maintained solution and service developers do not need to write additional code to include
+* It would not break parity between the libraries as the style would be computed in the browser
+
+### Negative Consequences
+
+* It does not provide an opt-out option, but if that is required the stylesheet could be removed from
+  `all-govuk-and-hmrc.scss` and added as a configurable optional script in `HmrcScripts` in `play-frontend-hmrc`
```
oscarduignan, line 49: services can have their own print stylesheet - or @media print rules in their stylesheet that overrides it if they need to - it would be a bit crap if they didn't already have any custom styles, but not that bad - could always even just have an inline style tag to save the extra new network request if no custom asset controller already.

another alternative we could add in future if really needed - would be have an extra class ".suppress-hmrc-print-styles" and only apply these display none if not within a container with that class - but I think it's probably not needed for now and probably YAGNI

oscarduignan (approved):

PR #338, opened 2024-01-11
PLATUI-2705: Upgrade to latest govuk-frontend v5.0.0

kyle-bowden:
Stuff that has happened in this PR

- IE Pollyfills have been removed
- html5shiv has been left in as decided by us during our spike investigation 
- forked components have been updated
- TimeiutDialog was using ActiveXObject for IE which we have decided to leave in place until DIAS lets us know if we should be supporting the TimeoutDialog in IE


package.json
```diff
@@ -106,7 +104,8 @@
     "stylelint": "^13.13.1",
     "stylelint-config-standard": "^20.0.0",
     "stylelint-scss": "^3.16.0",
-    "yargs": "^15.3.1"
+    "yargs": "^15.3.1",
+    "nodelist-foreach-polyfill": "^1.2.0"
```
oscarduignan: small one, but we can actually remove the polyfill now because [it was only IE that didn't have `NodeList.forEach`](https://caniuse.com/mdn-api_nodelist_foreach) 

just a random drive by skim since I left some tabs open - will review properly in morning!

tasks/gulp/compile-js.js
oscarduignan, line 1: probably lots of improvements we could snag from https://github.com/alphagov/govuk-frontend/blob/60b0a67bcd7cbb9c072c0fd573c30cdd5c2d31ba/packages/govuk-frontend/rollup.publish.config.mjs I think it's outside scope of this change probably - though maybe some bits we might want to do

package.json
kyle-bowden: Ah, I followed the TODO in the [common.js file](https://github.com/hmrc/hmrc-frontend/blob/main/src/common.js), is this not a valid approach?

src/components/header/template.test.js
```diff
@@ -247,15 +247,6 @@ describe('header', () => {
 
   describe('SVG logo', () => {
     const $ = render('header', {});
-    const $svg = $('.govuk-header__logotype-crown');
-
-    it('sets focusable="false" so that IE does not treat it as an interactive element', () => {
```
JoPintoPaul, line 252: This test has been deleted but it looks like we still have the SVG with focusable="false", should we keep the test? Or delete the attribute?

kyle-bowden: I replaced the SVG element with the one from the guidance on the govuk-frontend release notes. I am going to leave the test that tests for focusable="false" as this is what govuk recommends.

JoPintoPaul (approved):

oscarduignan (approved):

PR #339, opened 2024-01-24
PLATUI-2786: fix issue with hmrc components that import govuk components

oscarduignan:
The explanation in the README was a bit of sleuthing to make sure we were updating all the paths that needed to be changed and checking that the CSS path changes in 6.0.0 will work for services and prototypes

Still need to package this locally and test in a prototype, tested locally and it works

(tested by doing `npm run build` creating a new prototype, and in the prototype doing `npm install path/to/hmrc-frotend/package` and installing/starting prototype and using hmrcAddToList (which imports govuk component) and seeing it doesn't error anymore)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,60 @@ npm install
 
 For use in frontend microservices running on MDTP, please refer to the [README](https://www.github.com/hmrc/play-frontend-hmrc)
  in play-frontend-hmrc.
-    
+
+## Explanations
+
+### Relative `@import`s and SASS load paths
+
+Our SASS needs to be compilable in three different contexts:
+
+1. Within the library itself
+2. Within the node_modules folder of prototypes
+3. Within the folder that webjar assets get extracted to in tax services
+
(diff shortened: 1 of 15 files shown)


TimothyFothergill (approved):
aww yes

README.md
```diff
@@ -46,7 +46,60 @@ npm install
...
+We have
+an [automated test to ensure that we don't use an invalid relative path that would break this use case.](/tasks/gulp/__tests__/after-build-package.test.js#L140-L151)
+
+##### 3. Within the folder that webjar assets get extracted to in tax services
+
+The sbt-sassify plugin used by most tax services to compile SASS also has no way to add to the SASS load paths. Luckily,
+the subfolder structure that assets from webjars are extracted into matches what you'd find in node_modules, so the
+relative import paths that work for node_modules for prototypes will also work here (so long as you've got the
+govuk-frontend webjar in your project.)
+
```
ellamdav, line 102: 😍

PR #340, opened 2024-02-08
PLATUI-2823: Update govuk-frontend to v4.8.0 to include new Tudor cro…

JoPintoPaul:
…wn, updated header with feature flag, added new HMRC Crest Logo to banner and internal header with feature flag


src/components/internal-header/internal-header.yaml
```diff
@@ -20,6 +20,10 @@ params:
   type: string
   required: false
   description: The public path for the assets folder. If not provided it defaults to /assets
+- name: useDeprecatedCrown
+  type: boolean
+  required: false
+  description: A feature flag to use old St Edward’s Crown. If "false" or unset, defaults to the Tudor Crown
```
oscarduignan: @JoPintoPaul I think we need to [match the govuk-frontend pattern of defaulting to the st edwards crown](https://github.com/alphagov/govuk-frontend/blob/78ca753b4e67f4644dd1e3fe8fedab10c780d56d/packages/govuk-frontend/src/govuk/components/header/template.njk#L46) to start with - otherwise if we release this then if someone is pulls this into a play-nunjucks service for example then they'll immediately start using the tudor crown rather than having to enable it (for safety before the 17th)

oscarduignan: it's confusing because the idea then is that in play-frontend-hmrc we make components default to use the tudor crown unless there's some app config to turn it off - which we add to app-config-common before we release the play-frontend-hmrc version so that it can't go live until we remove that

JoPintoPaul: @oscarduignan On it.

backstop_data/bitmaps_reference/backstop_default_header_with_tudor_crown_example_0_document_1_tablet.png
oscarduignan, line 1: I can't remember if we pass the hmrc banner through or just have a param to enable it - I guess if we just have a param we should pass through the use tudor crown setting to it - because here it's got the tudor crown in the header and then the st edwards one in the hmrc banner

oscarduignan:
closing this - has been superseded

PR #344, opened 2024-02-23
Plat UI 2847

oscarduignan:
- [x] make sure I've got all the tudor crown stuff correctly 😵 (seems like it from the backstop tests)
- [x] flag with design resources and dias
  - should we be changing how the hmrc banner works to be in line govuk approach (no fallback ie8 image and path of svg all in one to avoid some a glitch they're documented in a comment)
  - new typographic scale coming in v6 that's available now to opt-into early to try out
  - for the header to work, I've had to put the extra things outside the header (which seems correct according to gds- but not sure how big of a change this will be for teams - something to talk about before merging)
- [x] does hmrc character count need any changes to support attributes / before and after html?
- [x] are there any header changes that need applying to our fork?
- [x] are there any footer changes that need applying to our fork?
- [x] rerun backstop tests and accept the changes
- [x] fix unit tests which svgs changing have broken
- [x] pull in latest npm audit fix and update changelog

Future stuff:
- [ ] refactor warnings to use govuk-font-size mixin
- [ ] there are some TODOs in hmrc header for possible improvements / bugs to fix in the future
- [ ] in play-frontend-hmrc we are reading "formGroup.classes" into "formGroupClasses" rather than an object, we now have extra params in a formGroup so we can deprecate formGroupClasses and add a formGroup param instead and use that if formGroupClasses is empty - I guess - this FormGroup param is shared amongst a few components


TimothyFothergill (dismissed):
wow so good

src/components/header/template.njk
```diff
@@ -1,105 +1,135 @@
...
-      {% if params.serviceName or params.navigation | length or params.signOutHref or params.languageToggle %}
-      <div class="govuk-header__content">
-      {% if params.serviceName %}
-        {% set serviceNameTag = 'a' if params.serviceUrl else 'span' %}
-        <{{ serviceNameTag }} {% if (serviceNameTag == 'a') %}href="{{ params.serviceUrl }}" {% endif %} class="hmrc-header__service-name {% if (serviceNameTag == 'a') %}hmrc-header__service-name--linked{% endif %}">
-          {{ params.serviceName }}
-        </{{ serviceNameTag }}>
+{%- set menuButtonText = 'Dewislen' if language == 'cy' else 'Menu' -%}
+
+{%- set _stEdwardsCrown %}
```
timsb, line 10: What's the significance of the underscore here? Is it a new naming style?

oscarduignan: apologies Tim, I missed this - the underscore is because I've copied the implementation from the govuk header and that's how they've done it- I'm not sure what the significance is for them (maybe that it's temporary / will go away soon - but not sure why that would be important to convey here because I don't think the variables would be accessible anywhere else)

oscarduignan:
bit of a last minute backtrack here, will let my commit describe what and why

https://github.com/hmrc/hmrc-frontend/pull/344/commits/82d690a58f63aedc796f003feeb43b5629a50b90

> In trying to create an example of the issue that I changed the structure to fix, I realised that the issue [I was seeing] came not from a govuk change but from changing back to follow the govuk convention [and not applying our patch to it - moving the header to a wrapper and changing govuk-header to a div]. Because of that, even though I still think it would be good to match GOV.UK - it's not worth making that change as part of this - especially because we still have the language select and sign out links within the header which mean we wouldn't be able to get rid of our fork all together.

src/components/header/_header.scss
oscarduignan, line 1: reverted because I found some services that are building the markup themselves (and there may be more) so for now it's not safe to remove without first getting these people to stop using this old markup

https://github.com/search?q=org%3Ahmrc+hmrc-header__service-name+path%3A%2F%5E*.scala.html%24%2F&type=code

src/components/banner/template.njk
```diff
@@ -1,18 +1,43 @@
...
-      <path d="M67.47,7.72a59.76,59.76,0,1,0,59.75,59.75A59.75,59.75,0,0,0,67.47,7.72m67.47,59.75A67.47,67.47,0,1,1,67.47,0a67.47,67.47,0,0,1,67.47,67.47" />
-    </svg>
+    {% if params.useTudorCrown | default(true) %}
+      <!--[if gt IE 8]><!-->
```
JoPintoPaul: Hi @oscarduignan , I see we've still got the IE8 fallback in here which is now removed from the `govuk-frontend` templates, shall we delete these fallbacks?

src/components/banner/template.test.js
```diff
@@ -27,5 +27,13 @@ describe('Internal Header', () => {
...
...mrc-organisation-logo > svg').html().trim()).toContain('<path d="M104.32,73.72,101,73.29c.94-2,1.51-3.22,1.51-3.22,0-.38,1.11-2.22,1.53-3.73a19.94,19.94,0,0,0,.61-3.89,15.05,15.05,0,0,0-3.68-10.06,12.21,12.21,0,0,0-9.18-4.18,17.46,17.46,0,0,0-3.68.44c-2.09.52-3.08.81-4.93,1.29a51.13,51.13,0,0,0-8.85,3.61,1.37,1.37,0,0,0-.73.91,2,2,0,0,0-.07.53L73.31,71.6H70.53V53.46a4.05,4.05,0,0,1,.18-1.14,3.37,3.37,0,0,1,1.43-1.71c.6-.39,1.21-.75,1.82-1.08a2.49,2.49,0,1,1,3.87-2.07c0,.12,0,.25,0,.37l.16-.06c.55-.19,1.16-.37,1.75-.55a2.49,2.49,0,1,1,3.71-2.16,2.66,2.66,0,0,1-.3,1.16l2.06-.56.36-.08a2.49,2.49,0,1,1,3.87-2.07A2.42,2.42,0,0,1,89,44.82h.06a16.43,16.43,0,0,1,2.72-.18,2.45,2.45,0,0,1-.48-1.45,2.49,2.49,0,1,1,5,0,2.46,2.46,0,0,1-.89,1.89A16.46,16.46,0,0,1,98.1,46a2.41,2.41,0,0,1-.12-.72,2.5,2.5,0,1,1,3,2.44,16.7,16.7,0,0,1,2.12,1.91v-.08a2.49,2.49,0,1,1,2.49,2.49,2.15,2.15,0,0,1-.68-.11,18.34,18.34,0,0,1,1.25,2.26,2.49,2.49,0,0,1,4.55,1.4,2.5,2.5,0,0,1-2.49,2.5,2.64,2.64,0,0,1-.8-.15,20.9,20.9,0,0,1,.35,2.1,2.52,2.52,0,0,1,1-.19,2.49,2.49,0,1,1,0,5,2.59,2.59,0,0,1-1.07-.25c0,.26-.08.54-.11.82s-.14.55-.21.78a2.49,2.49,0,0,1-.32,5,2.39,2.39,0,0,1-1.39-.43c-.34.8-.78,1.83-1.27,3"></path>');
```
JoPintoPaul: Can probably trim the match down to not include the whole SVG, but this is just a suggestion.

CHANGELOG.md
```diff
@@ -25,6 +33,38 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Updated to work with govuk-frontend v5
 
+## [5.67.0] - 2024-02-19
```
ellamdav, line 36: 🙌

src/components/internal-header/template.njk
```diff
@@ -2,20 +2,45 @@
...
-                                <path d="M67.47,7.72a59.76,59.76,0,1,0,59.75,59.75A59.75,59.75,0,0,0,67.47,7.72m67.47,59.75A67.47,67.47,0,1,1,67.47,0a67.47,67.47,0,0,1,67.47,67.47" />
-                        </svg>
-                        <span class="hmrc-internal-header__logo-text">{% if params.language == 'cy' %}
+					{% if params.useTudorCrown | default(true) %}
+            <!--[if gt IE 8]><!-->
```
JoPintoPaul: Again, spotted that the IE8 stuff has crept back in.

src/components/footer/template.njk
```diff
@@ -1,89 +1,35 @@
...
+  >{% if(params.language == 'cy') %}y Drwydded Llywodraeth Agored{% else %}Open Government Licence{% endif %} v3.0</a>, {% if(params.language == 'cy') %}oni nodir yn wahanol{% else %}except where otherwise stated{% endif %}
+{%- endset %}
+
+{% set copyrightHtml -%}
+© {% if(params.language == 'cy') %}Hawlfraint y Goron{% else %}Crown copyright{% endif %}
+{%- endset %}
+
+{% set defaultMetaVisuallyHiddenTitle = "Cysylltiadau cymorth" if params.language == "cy" else "Support links" %}
+
+{{ govukFooter({
```
ellamdav, line 18: 🙌

src/components/footer/template.test.js
```diff
@@ -14,6 +14,13 @@ describe('footer', () => {
     return (text || '').trim().replace(/\s+/g, ' ');
   }
 
+  it('should match the output of govuk footer when displayed in english', async () => {
```
ellamdav, line 17: ❤️

src/components/header/template.test.js
```diff
@@ -192,76 +190,26 @@ describe('header', () => {
...
-      it('is invisible to modern browsers', () => {
-        expect($fallbackImage.length).toEqual(0);
+        const $button = $('.govuk-header__menu-button');
+        expect($button.text().trim()).toEqual('Dewislen');
       });
     });
   });
 
+  // TODO add tests for sign-out-links
+  // TODO add tests for language-select-links
```
ellamdav, line 211: were these for now, or just future tech debt?

src/components/banner/template.njk
oscarduignan: I'll remove them - I'm not sure how I added them back in - I even left a comment about if we need them but they were already removed and I've added them back 🤦‍♂️

src/components/header/template.test.js
oscarduignan: I put them on the tech debt board for future - I have some questions about how these bits work so I didn't want to enshrine right now the current implementation too much

src/components/banner/template.njk
oscarduignan: done!

src/components/banner/template.test.js
oscarduignan: good idea, done!

src/components/header/template.njk
timsb: No problem, was just curious. I thought there'd be a reason even if only to keep consistency from govuk-frontend

src/components/internal-header/template.njk
oscarduignan: done!

JoPintoPaul (approved):

PR #348, opened 2024-03-12
NOJIRA - Setup github actions

TimothyFothergill:
# Purpose of PR
Setting up GitHub actions on hmrc-frontend to run npm audit fix and raise a PR

github action will now take place on the 50th minute of every hour - This will be changed to once per day once tested. It can't automatically merge the PR it raises, so safe to see what happens.

### DIFF
--- a/.github/workflows/dependency-updates.yml
+++ b/.github/workflows/dependency-updates.yml
@@ -8,7 +8,7 @@ on:
     branches:
       - main
   schedule:
-    - cron: "55 * * * *"
+    - cron: "50 * * * *"
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-updates.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  schedule:
+    - cron: "50 * * * *"
```
tmc-ee, line 11: ~Is this ran too frequently?~ Ignore this, explained in Slack that this is temporary 👍

tmc-ee (approved):

.github/workflows/dependency-updates.yml
TimothyFothergill: For anyone who comes to this in future: Tom's absolutely right. The goal would be to then reduce the frequency to once per day.

PR #360, opened 2024-03-27
PLATUI-2829: Revert change to retain the back link even when the referrer is empty

kyle-bowden:


src/components/back-link-helper/back-link-helper.js
```diff
@@ -5,27 +5,14 @@ function BackLinkHelper($module, window, document) {
...
-    } catch (err) {
-      return false;
-    }
-  };
-
   // do nothing if History API is absent
   if (this.window.history) {
-    // hide the backlink if the referrer is on a different domain
-    if (isReferrerExternal()) {
+    // TODO: Appsec have not yet given a clear answer on if we need to do the same domain check
```
JoPintoPaul: Do we want this TODO to reference today's discussion? There's the AppSec input, but also the UR consideration.

JoPintoPaul (dismissed):

src/components/back-link-helper/back-link-helper.js
```diff
@@ -5,27 +5,20 @@ function BackLinkHelper($module, window, document) {
...
-    if (isReferrerExternal()) {
+    // eslint-disable-next-line max-len
+    /* TODO: It remains unclear whether a check for the same domain is necessary for security reasons.
+       There may be user research suggesting considerations regarding the visibility of the
+       back link on refresh.
+       Currently, a page refresh sets the referer to empty, leading to the back link being hidden
+       under our existing logic.
+     */
+    // eslint-disable-next-line max-len
+    const referrerNotOnSameDomain = () => this.document.referrer.indexOf(this.window.location.host) === -1;
```
oscarduignan: will this cause a js error if document.referrer is empty? I don't know if it will be "" or undefined, but if it can be undefined then indexOf might cause an error

kyle-bowden: According to the Mozilla documentation, the referer will always be set to an empty string if it does not have a value.
(but its always best to have a safeguard, so i can put that in)
https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer

oscarduignan: that is good enough for me to be honest, can I use shows even old IE seems to support, so not sure why it was coded like that in old version, probably just bit more defensive coding than needed, https://caniuse.com/mdn-api_document_referrer

oscarduignan (dismissed):

oscarduignan (approved):

PR #363, opened 2024-04-02
PLATUI-2908: Added outline-offset to account menu for accessibility

JoPintoPaul:


package-lock.json
```diff
@@ -1928,21 +1928,6 @@
           "integrity": "sha512-bN798gFfQX+viw3R7yrGWRqnrN2oRkEkUjjl4JNn4E8GxxbjtG3FbrEIIY3l8/hrwUwIeCZvi4QuOTP4MErVug==",
           "dev": true
         },
-        "color-convert": {
```
oscarduignan, line 1931: I still have no idea what causes this shuffling around of dependencies sometimes 😵

oscarduignan:
looks good to me assuming ok with DIAS

took a quick look at https://github.com/hmrc/sca-wrapper and at a glance it looks like there wouldn't be any problems with the css with how they build the markup https://github.com/hmrc/sca-wrapper/blob/92ebaed04c62354a55fd39d2fd8b8e1d6ca7a376/play-30/src/main/twirl/uk/gov/hmrc/sca/views/PtaMenuItemLeft.scala.html#L24 and cant' see any custom css that could conflic

really brittle and easy to break, but not sure what we could do to make the dependency clearer in the future

ideally wouldn't be a dependency but don't think that will ever happen

oscarduignan (approved):

PR #364, opened 2024-04-05
PLATUI-2941: Update hmrc-frontend to govuk-frontend v5.3.0

kyle-bowden:


src/components/footer/template.test.js
```diff
@@ -28,13 +28,6 @@ describe('footer', () => {
     expect(results).toHaveNoViolations();
   });
 
-  it('entire component must have a role of `contentinfo`', () => {
```
kyle-bowden, line 31: This test was removed because it was checking the role attribute which has now been [removed from the govukFooter component](https://github.com/alphagov/govuk-frontend/commit/badceea3a041cd8ae798197414214ecb64057770)

TimothyFothergill (dismissed):
upgrade

src/components/footer/template.test.js
ellamdav: We should remove [this section](https://github.com/hmrc/hmrc-frontend/blob/bf06fea63d98c0b81aabfdd1cde260a49264f898/src/components/footer/footer.yaml#L107-L113) of the footer yaml if it's no longer relevant

ellamdav: Should we also follow GOVUK's lead and remove `role` from [our own forked components](https://github.com/search?q=repo%3Ahmrc%2Fhmrc-frontend%20role&type=code)?

ellamdav:
Think we also need to remove references to the `role` from [header.yaml](https://github.com/hmrc/hmrc-frontend/blob/main/src/components/header/header.yaml#L143-L145) and [internal-header.yaml](https://github.com/hmrc/hmrc-frontend/blob/main/src/components/internal-header/internal-header.yaml#L44-L46) - though I'm also not sure who these `accessibilityCriteria` are for, or if the rest of them are up-to-date.

ellamdav (approved):

PR #374, opened 2024-06-05
⛔️ DO NOT MERGE ⛔️ stop using deprecated 14px typography style in account menu and it's notification badge

oscarduignan:
⛔️ DO NOT MERGE ⛔️ 

This is an example to use to discuss the impact of change / what we should do about it with PTA and Design Resources - as it might need some coordination and the actual change might be different - or contain additional tweaks to make new font size fit (at the least)

https://github.com/alphagov/govuk-design-system/issues/2289


oscarduignan:
copying a summary from some initial slack discussion, sounds like maybe rather than going to 16px and then having a change to coordinate across lots of services, we might just inline the current behaviour at 14px so that we aren't impacted by the deprecation, then not worry about maintainability of that because the new navigation patterns from gov.uk are on the horizon and likely we'd be better investing time in rolling that out when it's available rather than updating this pattern

oscarduignan:
we have now inlined the compiled css for this, so we aren't impacted by the deprecation- and eventually this menu will be replaced by new service navigation component

PR #379, opened 2024-06-20
PLATUI-2603: make serviceUrl optional for internal header

oscarduignan:
as part of this we discovered that we need to update how the font-sizes are set to use rem rather than px, and introduce hover styles (which @timsb is currently working on separately)

creating this without converting to use rem in case we want to bundle that into Tim's change

remaining tasks:
- [x] add a unit test to check the markup output is correct


src/components/internal-header/template.njk
oscarduignan, line 1: whitespace changes, there was inconsistent indentation between 2 spaces and 4 spaces

src/components/internal-header/template.njk
```diff
@@ -1,48 +1,52 @@
...
+    </div>
+    {%- if params.serviceName -%}
+    <div class="hmrc-internal-header__service-name">
+      {% if params.serviceUrl %}
+        <a href="{{ params.serviceUrl }}" class="hmrc-internal-header__link">
+          {{ params.serviceName }}
+        </a>
+      {% else %}
+        {{ params.serviceName }}
+      {% endif %}
```
oscarduignan, line 48: this is the only actual change in this file besides whitespace

src/components/internal-header/template.njk
oscarduignan: suggest setting "hide whitespace" when viewing this PR!

oscarduignan: (I don't always recommend it because html is whitespace sensitive in inline contexts)

TimothyFothergill (approved):
nailed it

PR #383, opened 2024-06-27
PLATUI-3119: Fix navigation styling in header component

kyle-bowden:


src/components/header/header.yaml
```diff
@@ -476,3 +476,120 @@ examples:
     menuButtonText: 'Custom button text'
     menuButtonLabel: 'Custom button label'
     navigationLabel: 'Custom navigation label'
+- name: with custom navigation items and a sign out link
```
oscarduignan, line 490: might be worth having two extra test cases in VRT (or trying it out to check)
- one with less nav items
- and one with enough to have multiple lines

oscarduignan: I don't actually know how multiple lines will look

oscarduignan: it might also be good to split this into two commits, so we can see the changes before and after

basically could just undo the _header.css change on this commit and redo it in a new commit - where one before has the addition of the new backstop scenarios (with issue visible) and then new commit has the fixes and we can compare the changes between them

kyle-bowden: I have added two more examples

_with service name and one custom navigation item and a sign out link
with service name and lots of custom navigation items and a sign out link_

backstop_data/bitmaps_reference/backstop_default_header_with_service_name_and_custom_navigation_items_and_a_sign_out_link_example_0_document_0_phone.png
oscarduignan, line 1: @kyle-bowden so the fix we've got so far stops the sign out and translation links from being able to appear on the same line as the service name when laid out on mobile

this is a regression that people might raise as a new bug I think - I wouldn't want any of us to approve a change like this

and rather than try to see if all the stakeholders are ok with this, maybe we can tweak the fix some more so we don't get this happening on mobile

I guess also, I'm assuming it's not changing desktop from:

```
[ GOV.UK ] [ Service Name                                    ] [ Sign Out ]
```

to:

```
[ GOV.UK ] [ Service Name                                                 ]
                                                               [ Sign Out ]
```
because there's no change to those screenshots (and I'm assuming somewhat that there are ones with service name / sign out and no nav - and no service name and only sign out link - might be worth checking those just in case 😅 )

oscarduignan: part of me is actually wondering if we should merge the extra test cases into main first- but not update package.json and let it just fail the build - then our release will get this nice diff of changes too

PR #390, opened 2024-07-12
Proof of concept: migrate to dart-sass

oscarduignan:
The key bit of this proof of concept is that by bodging the includePaths we've made it so that we get the behaviour we had before from libsass which means we don't need to change the relative imports we have in our sass paths that will require sass load paths be updated by consumers. Where some of our consumers are using different ways of loading packages and they won't always be in node_modules folder, and some don't have the option of updating their load paths (govuk prototype kit) users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,8 +42,8 @@
   },
   "homepage": "https://github.com/hmrc/hmrc-frontend#readme",
   "dependencies": {
-    "govuk-frontend": "^5.3.0",
-    "accessible-autocomplete": "^2.0.4"
+    "accessible-autocomplete": "^2.0.4",
+    "govuk-frontend": "^5.3.0"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.21.5",
@@ -100,6 +100,7 @@
     "rollup-plugin-babel": "^4.4.0",
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-node-resolve": "^5.2.0",
+    "sass": "^1.77.8",
     "standard": "^17.0.0",
     "stylelint": "^13.13.1",
     "stylelint-config-standard": "^20.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -100,6 +100,7 @@
     "rollup-plugin-babel": "^4.4.0",
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-node-resolve": "^5.2.0",
+    "sass": "^1.77.8",
```
oscarduignan, line 103: would also need to remove node-sass, but didn't need to for the proof of concept

tasks/gulp/compile-scss.js
```diff
@@ -22,7 +22,7 @@ gulp.task('scss:compile-all-govuk-and-hmrc', () => gulp.src(`${configPaths.src}a
   .pipe(plumber(errorHandler))
   .pipe(sourcemaps.init())
   .pipe(sass({
-    includePaths: ['node_modules'],
+    includePaths: ['node_modules', 'node_modules/govuk-frontend/dist', 'node_modules/govuk-frontend/dist/govuk/components'],
```
oscarduignan, line 25: we have two types of relative imports at the moment `../../govuk-frontend` and `../../../../govuk-frontend` which - I think govuk-frontend gets around issues with load paths by not importing any css from node_modules in their css - they just have their own css / stuff vendored in that's a single file

the way this works, we don't actually want to use any css from either of those new paths, we just want to be able to relatively traverse to places in govuk-frontend from hmrc-frontend css when we're compiling it when the hmrc-frontend css is not inside node_modules - all our consumers compile the hmrc css in the context of hmrc-frontend being inside a folder (like node_modules or lib in the case of webjars) where govuk-frontend will be also in that folder, so relative traversals for them work

oscarduignan:
superseded by https://github.com/hmrc/hmrc-frontend/pull/395 I'm closing this because future collaboration can happen in the other PR

PR #391, opened 2024-07-17
PLATUI-3035 remove unused report-technical-problem url param

TimothyFothergill:
# Purpose of PR
- To remove an unused url parameter from the report-technical-problem component
- Update all tests

### DIFF
--- a/src/components/report-technical-issue/template.njk
+++ b/src/components/report-technical-issue/template.njk
@@ -1,9 +1,9 @@
 {% set language %}{% if params.language == 'cy' %}cy{% else %}en{% endif %}{% endset %}
 {% set serviceId = params.serviceId or params.serviceCode %}
 {% set urlParametersBuilder %}
-  {% if serviceId %}?service={{ serviceId | urlencode }}{% endif %}
+  {% if params.referrerUrl or serviceId %}?{% endif %}
+  {% if serviceId %}service={{ serviceId | urlencode }}{% endif %}
   {% if params.referrerUrl and serviceId %}&amp;{% endif %}
-  {% if params.referrerUrl and not serviceId %}?{% endif %}
   {% if params.referrerUrl %}referrerUrl={{ params.referrerUrl | urlencode }}{% endif %}
 {% endset %}
 <a lang="{{ language }}" hreflang="{{ language }}"
--- a/src/components/report-technical-issue/template.njk
+++ b/src/components/report-technical-issue/template.njk
@@ -4,5 +4,5 @@
   class="govuk-link hmrc-report-technical-issue {{ params.classes }}"
   {% if params.referrerUrl %} rel="noreferrer noopener"{% endif %}
   target="_blank"
-  href="{{ params.baseUrl }}/contact/report-technical-problem?newTab=true{% if serviceId %}&amp;service={{ serviceId | urlencode }}{% endif %}{% if params.referrerUrl %}&amp;referrerUrl={{ params.referrerUrl | urlencode }}{% endif %}"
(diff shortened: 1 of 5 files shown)


src/components/report-technical-issue/template.njk
```diff
@@ -4,5 +4,5 @@
   class="govuk-link hmrc-report-technical-issue {{ params.classes }}"
   {% if params.referrerUrl %} rel="noreferrer noopener"{% endif %}
   target="_blank"
-  href="{{ params.baseUrl }}/contact/report-technical-problem?newTab=true{% if serviceId %}&amp;service={{ serviceId | urlencode }}{% endif %}{% if params.referrerUrl %}&amp;referrerUrl={{ params.referrerUrl | urlencode }}{% endif %}"
+  href="{{ params.baseUrl }}/contact/report-technical-problem{% if serviceId or params.referrerUrl %}?{% endif %}{% if serviceId %}service={{ serviceId | urlencode }}{% endif %}{% if params.referrerUrl %}&amp;referrerUrl={{ params.referrerUrl | urlencode }}{% endif %}"
```
oscarduignan: ah, I think we might need to tweak this

if you have no service id but you do have a referrer url then you'll get a url like `?&amp;referrerUrl` which might be ok - I'm not sure - but should be `?referrerUrl` in that case

oscarduignan: logic is something like

```
if serviceId then "?serviceId=$serviceId"

if referrerUrl then
  (if serviceId then "&amp;" else "?")
    + "referrerUrl=$referrerUrl"
```

but might be nicer ways of writing it 😅

TimothyFothergill:
```nunjucks 
  {% set urlBuilder %}
  {% if serviceId %}?service={{ serviceId | urlencode }}{% endif %}
  {% if params.referrerUrl and serviceId %}&amp;{% endif %}
  {% if params.referrerUrl and not serviceId %}?{% endif %}
  {% if params.referrerUrl %}referrerUrl={{ params.referrerUrl | urlencode }}{% endif %}
  {% endset %}
```
```nunjucks
  href="{{ params.baseUrl }}/contact/report-technical-problem{{urlBuilder}}"
```
This seems to work with a few tweaks to the tests to accommodate. Separating it out from the href param feels less confusing. Thoughts?

TimothyFothergill: Will update `urlBuilder` to `urlParametersBuilder`

src/components/report-technical-issue/template.njk
```diff
@@ -1,8 +1,14 @@
 {% set language %}{% if params.language == 'cy' %}cy{% else %}en{% endif %}{% endset %}
 {% set serviceId = params.serviceId or params.serviceCode %}
+{% set urlParametersBuilder %}
+  {% if serviceId %}?service={{ serviceId | urlencode }}{% endif %}
+  {% if params.referrerUrl and serviceId %}&amp;{% endif %}
+  {% if params.referrerUrl and not serviceId %}?{% endif %}
```
TimothyFothergill: Technically, I guess the ? setter could be the first part, where we just look for `serviceId` or `params.referrerUrl`.

src/components/report-technical-issue/template.njk
```diff
@@ -1,8 +1,14 @@
 {% set language %}{% if params.language == 'cy' %}cy{% else %}en{% endif %}{% endset %}
 {% set serviceId = params.serviceId or params.serviceCode %}
+{% set urlParametersBuilder %}
+  {% if params.referrerUrl or serviceId %}?{% endif %}
+  {% if serviceId %}service={{ serviceId | urlencode }}{% endif %}
+  {% if params.referrerUrl and serviceId %}&amp;{% endif %}
+  {% if params.referrerUrl %}referrerUrl={{ params.referrerUrl | urlencode }}{% endif %}
+{% endset %}
```
oscarduignan, line 8: and the output doesn't get any extra spaces / newlins in it? 🤯 (seems not from the tests) beautiful!

oscarduignan (approved):

PR #399, opened 2024-08-22
PlatUI-3121 replace mixin with css

goodvibes:
https://jira.tools.tax.service.gov.uk/browse/PLATUI-3121

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "6.27.0",
+  "version": "6.28.0",
   "description": "Design patterns for HMRC frontends",
   "scripts": {
     "start": "gulp dev",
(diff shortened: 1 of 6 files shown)


tasks/gulp/compile-scss.js
```diff
@@ -27,7 +27,7 @@ gulp.task('scss:compile-all-govuk-and-hmrc', () => gulp.src(`${configPaths.src}a
   // minify css add vendor prefixes and normalize to compiled css
   .pipe(postcss([
     autoprefixer,
-    cssnano,
+    // cssnano,
```
oscarduignan: @goodvibes we need to uncomment this so that the css gets minified

goodvibes: Ohhhhhhh

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hmrc-frontend",
-  "version": "6.28.0",
+  "version": "6.29.0",
```
oscarduignan, line 3: if you set this to 6.28.0 and then run
```
npm version minor --git-tag-version=false
```
it will update this again and update the version number in package-lock.json which also needs to be updated - (suggest doing like this as I can't remember all the places in package-lock.json it is, think there migth be more than one so manually updating it might leave it partly incorrect)

oscarduignan: we've set git-tag-version=false in .npmrc but just in case it doesn't apply I've included it here

goodvibes: Done 👍

oscarduignan (approved):

PR #403, opened 2024-09-13
PLATUI-3224 update jest

oscarduignan:
update jest, refactored some test config as a result, had some broken tests with the timeout dialog which were hard to unpick/isolate with jsdom so I've refactored a lot of them into the browser-based tests instead, seems like we don't need --runInBand (run tests in sequence) so they should run faster now, but we'll see if that holds up in jenkins.

- [x] I think I've still got a bit of tidying up to do before this is totally ready for review.

most of this is moving around test coverage and trying to simplify test code so it's easier to maintain in future 🤞


config/jest-setup.js
```diff
@@ -1,5 +0,0 @@
-/* eslint-env jest */
-
-const { toHaveNoViolations } = require('jest-axe');
-
-expect.extend(toHaveNoViolations);
```
oscarduignan, line 5: didn't seem to be needed

.eslintrc.json
```diff
@@ -4,7 +4,15 @@
...
-    "browser": true
+    "browser": true,
+    "jest": true
+  },
+  "globals": {
+    "page": true,
+    "browser": true,
+    "context": true,
+    "puppeteerConfig": true,
+    "jestPuppeteer": true
```
oscarduignan, line 15: stops ide warnings about globals

jest.config.mjs
```diff
@@ -0,0 +1,46 @@
...
+    {
+      displayName: 'browser tests',
+      preset: 'jest-puppeteer',
+      clearMocks: true,
+      setupFilesAfterEnv: ['expect-puppeteer', '<rootDir>/lib/browser-tests/jest-setup.js'],
+      testMatch: [
+        '**/*browser.test.js',
+      ],
+    },
+  ],
```
oscarduignan, line 41: previously the environment of a test was configured by comments at the top of test files, there were some limits of what custom options you could specify that way so instead I've opted to rename tests with different suffixes. This matches the approach that govuk-frontend uses too https://github.com/alphagov/govuk-frontend/blob/main/jest.config.mjs

jest.config.mjs
```diff
@@ -0,0 +1,46 @@
...
+
+  modulePathIgnorePatterns: ['<rootDir>/package'],
```
oscarduignan, line 43: stops us getting a warning from jest haste map

lib/browser-tests/puppeteer-helpers.js
```diff
@@ -0,0 +1,67 @@
...
+const pathToSinonjsBrowserPkgSrc = path.resolve(__dirname, '../../node_modules/sinon/pkg/sinon.js');
+
+export async function useFakeTimers(page, config = {}) {
+  await page.evaluateOnNewDocument(`
+    if (!window.clock) {
+      ${fs.readFileSync(pathToSinonjsBrowserPkgSrc, 'utf8')}
+      window.clock = sinon.useFakeTimers(${JSON.stringify(config)})
+    }
+  `);
+}
```
oscarduignan, line 20: we were using sinon before - but a part of it only, and to do that we had to compile the extracted part from the browser, I discovered that if we just use the whole sinon library, it's npm package has a compiled js file ready to use in the browser, so no more need for that extra build step in our tests

lib/browser-tests/puppeteer-helpers.js
```diff
@@ -0,0 +1,67 @@
...
+  `;
+}
+
+export async function render(page, body) {
+  await page.setRequestInterception(true);
+  page.once('request', (req) => {
+    req.respond({ contentType: 'text/html', body });
+    page.setRequestInterception(false);
+  });
+  await page.goto('http://localhost:3000/'); // if ever flaky, waitUntil networkidle0 (js loaded)
```
oscarduignan: some of the tidying up needed is to replace anywhere I've used hardcoded port with the port from config

src/all.jsdom.test.js
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    it('should initialise each matching element found on the page', () => {
+      addBackLinkToPage();
+      addBackLinkToPage();
+      HMRCFrontend.initAll();
+      expect(BackLinkHelper.prototype.init).toHaveBeenCalledTimes(2);
+    });
+  });
+});
```
oscarduignan, line 36: this replaces src/all.test.js it's the same thing renamed and refactored but changed enough so that it's not picked up as a rename

package.json
```diff
@@ -111,17 +115,5 @@
...
-      "./config/jest-setup.js"
-    ],
-    "snapshotSerializers": [
-      "jest-serializer-html"
-    ],
-    "testEnvironment": "jsdom",
-    "testRunner": "jest-jasmine2",
-    "globalSetup": "./lib/puppeteer/setup.js",
-    "globalTeardown": "./lib/puppeteer/teardown.js"
-  }
```
oscarduignan, line 126: now configured in jest.config.mjs

package.json
```diff
@@ -100,6 +103,7 @@
     "rollup-plugin-babel": "^4.4.0",
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-node-resolve": "^5.2.0",
+    "sinon": "^18.0.1",
```
oscarduignan, line 104: probably some I can remove from package.json

src/components/currency-input/__snapshots__/jsdom.test.js.snap
oscarduignan, line 1: I think something is up with these currency component snapshots- doesn't seem to be actually rendering the component anymore - can't remember if that's because of changes to currency input which need to be reflected in tests (do we need them anymore? future tech debt ticket to be created)

src/components/timeout-dialog/timeout-dialog.jsdom.test.js
oscarduignan, line 1: I have left some tests here that were checking things like argument validation and message overrides - but moved the behavioural stuff into browser tests

src/components/timeout-dialog/timeout-dialog.browser.test.js
oscarduignan, line 1: this is where I extracted lots of timeout dialog tests into

src/components/timeout-dialog/timeout-dialog.browser.test.js
```diff
@@ -0,0 +1,408 @@
...
+    await background.bringToFront();
+    await expect(background).not.toShowTimeoutDialog();
+    await clockTickSeconds(background, 10);
+    await expect(background).toShowTimeoutDialog();
+    expect(await visualCountdownFrom(background))
+      .toBe('For your security, we will sign you out in 5 seconds.');
+    await session.close();
+  });
+
+  // TODO because currently it would actually log you out straight away, this is an edge case
```
oscarduignan, line 380: might be an edge case that in practice can't happen (I don't know if session activity could be delayed like this, so left this as a skipped test to show how it currently behaves)

src/components/timeout-dialog/timeout-multiple-tabs.browser.test.js
oscarduignan, line 1: another one where it already existed but renamed and refactored enough that it thinks it's new file

src/components/user-research-banner/user-research-banner.browser.test.js
```diff
@@ -1,25 +1,14 @@
...
- * @jest-environment ./lib/puppeteer/environment.js
- */
-/* eslint-env jest */
-
 const configPaths = require('../../../config/paths.json');
 
 describe('/components/user-research-banner', () => {
-  let page;
-  const url = `http://localhost:${configPaths.ports.test}/components/user-research-banner/default/preview`;
+  const url = `http://localhost:${configPaths.ports.app}/components/user-research-banner/default/preview`;
```
oscarduignan: I didn't think we needed a separate test port - we can just reuse the app one - and the jest can skip spinning up it's own dev server if ones already running

src/components/user-research-banner/user-research-banner.browser.test.js
```diff
@@ -28,10 +17,6 @@ describe('/components/user-research-banner', () => {
     await page.deleteCookie({ name: 'mdtpurr' });
   });
 
-  afterAll(async () => {
-    await page.close();
-  });
```
oscarduignan, line 33: in the browser-test jest-setup it resets the page before each test now to save stuff like this being duplicated around

tasks/gulp/backstop.js
```diff
@@ -26,7 +26,12 @@ gulp.task('backstop:approve', () => runBackstop('approve'));
 
 gulp.task('backstop-test', async () => {
   console.log(chalk.green('\nStart server for visual regression testing'));
-  const server = await appListen(port);
+
+  const server = new Promise((resolve) => {
```
oscarduignan: just seen this needs an await - which makes me think I haven't run backstop tests 👌

backstop_data/bitmaps_reference/backstop_default_timeout-dialog_timeout-warnings-not-synchronised_example_0_document_0_phone.png
oscarduignan, line 1: wanted a new example for some tests with specific timeout length, could have refactored it to not use the examples  and just render html into the test - which would be better but 🤷‍♂️

lib/browser-tests/puppeteer-helpers.js
```diff
@@ -0,0 +1,67 @@
...
+export async function clockTick(page, ms) {
+  // eslint-disable-next-line no-shadow
+  await page.evaluate((ms) => window.clock.tick(ms), ms);
+}
+
+export async function clockTickSeconds(page, s) {
+  await clockTick(page, s * 1000);
+}
+
+const preloadGovukFonts = fs.readdirSync(
```
JoPintoPaul, line 31: Is this so that the test is not slowed / flaky due to slow font loading?

src/components/timeout-dialog/timeout-dialog.browser.test.js
```diff
@@ -0,0 +1,408 @@
...
+      data-timeout-url="/timeout-reached"
+      data-sign-out-url="/signed-out-early"
+    `);
+    await clockTickSeconds(page, 800);
+    await expect(page).toClick('a', { text: 'Sign out' });
+    await page.waitForNavigation({ timeout: 500 });
+    await expect(page.url()).toMatch('/signed-out-early');
+  });
+
+  // TODO at the moment it doesn't, should we change test or implementation?
```
JoPintoPaul, line 184: This is an interesting one, worth capturing somewhere?

lib/browser-tests/puppeteer-helpers.js
oscarduignan: yup, for some of the timeout dialog tests there isn't any text rendered using the gov fonts on page load, only when the dialog triggers, and I guess css fonts are deferred by default so that's when the request is made to load them. Some of the timeout dialog tests intercept timeout / sign-out request to simulate a slow response and it was triggering on the fonts, so I copied over the preload code from the example app that's used for visual regression tests where we were getting some screenshots without fonts https://github.com/hmrc/hmrc-frontend/blob/5d05b9cb9745d51a088140277b16734206dad4cb/app/views/layouts/includes/head.html#L7

src/components/timeout-dialog/timeout-dialog.browser.test.js
oscarduignan: yup, could probably be a tech debt ticket, I'll create one and link it here

src/components/back-link-helper/back-link-helper.yaml
oscarduignan, line 1: I don't think we need VRT of the examples for this - just one example with the correct attribute set is enough

src/components/back-link-helper/back-link-helper.nojs.test.js
oscarduignan, line 1: was actually renamed and refactored, see above file

src/components/back-link-helper/back-link-helper.browser.test.js
oscarduignan, line 1: refactored the vrt examples into the browser tests, behaviour is a bit clearer like this I think, and with the examples we get extra screenshots being taken that we don't need

package.json
oscarduignan: removed sinonjs/fake-timers - we get this from sinon now instead

oscarduignan: removed browserify - we now use sinon that has js prepackaged for browser

JoPintoPaul (approved):

PR #405, opened 2024-10-01
PLATUI-3224: update dependencies and .nsprc

oscarduignan:
reviewing some security advisories, updating deps, removing exclusions we no longer need, and adding one new one that we seem not to be impacted by based on our usage


JoPintoPaul (dismissed):

oscarduignan:
something in the docker container for backstop 6.3.x (we were on backstop 6.1.x) changed the text rendering, so there are a lot of differences, I checked that it wasn't the fonts not loading, by changing the govuk font family to introduce serif before and after and rerunning the tests different times to check that transport font was actually being used - hopefully it still runs ok on jenkins... I'm not certain

oscarduignan:
this isn't going to be fun to debug, so jenkins failed and local tests failed after updating to 6.3.x of backstopjs - so the hope was to just accept the changes locally after verifying they were ok and it would also pass on build, even though build doesn't actually use the docker container, however it seems like there are still some failing tests

>        report | Test completed...
>        report | 357 Passed
>        report | 26 Failed

oscarduignan:
some odd differences, where stuff is different between renderings of different examples of the same component, will have to investigate the examples, but these ones might be around whitespace / block or inline context related things, so might require some change to the css to resolve. I might need to reverse the backstopjs update and create a tech debt ticket to investigate

CleanShot 2024-10-03 at 13 54 52@2x

oscarduignan:
looks like tests are also flakey, because I have gotten two failures rerunning the tests locally

CleanShot 2024-10-03 at 14 04 05@2x

oscarduignan:
possibly a timing change with backstop, based on this local failure

CleanShot 2024-10-03 at 14 05 35@2x

oscarduignan:
yup, seems like going past backstopjs 6.1.x we get flakey test results, not sure what to do about that but a bigger task than version bumps I was doing, so will turn that into tech debt

TimothyFothergill (approved):
lgtm

PR #406, opened 2024-10-03
PBD-4732 TEST DO NOT MERGE

PTMGitHub:
* Want to trigger an install of a non-standard nodejs version to test how the PR builder gets on when nodejs.org is blocked.


PTMGitHub:
Testing done. not needed

PR #415, opened 2024-11-18
PLATUI-3278 bump accessible-autocomplete

TimothyFothergill:
# Purpose of PR
- Bump accessible-autocomplete to `latest` (`3.0.1`)
- Expected test updates
- Update CHANGELOG


backstop_data/bitmaps_reference/backstop_default_accessible-autocomplete_default_example_0_document_0_phone.png
oscarduignan, line 1: @TimothyFothergill do we know where the extra padding comes from? Like can we include a link to show it intentional rather than accidental

Additionally, and it's maybe a stretch goal for this that we don't have to do, but I was wondering if it was worth adding in a test that loads the autocomplete with adam's patch and checks there are no js errors - could add it as an example, and then have a visual regression test that it looks ok, and a browser test that opens that example and checks browser console has no errors in it

PR #416, opened 2024-11-18
PLATUI-3278 bump accessible-autocomplete

TimothyFothergill:
# Purpose of PR
- Bump `accessible-autocomplete` to `latest` (`3.0.1`)
- Update `CHANGELOG` and provide a link to the updated styles
  - Note: There are some slight style changes provided by GDS in `3.0.0`.


oscarduignan:
aha, just clocked that it's only on mobile that it's changed size quite a bit - so seems an intentional change to me in accessible autocomplete and not any unintended interaction with our custom styles (checked the accessible autocomplete example site and changed width to simulate being on mobile and then it matches the look of the updated examples 👌)

oscarduignan (approved):

PR #417, opened 2024-11-27
PLATUI-3378 add pega domains to back link

goodvibes:
https://jira.tools.tax.service.gov.uk/browse/PLATUI-3378

### DIFF
--- a/src/components/back-link-helper/back-link-helper.js
+++ b/src/components/back-link-helper/back-link-helper.js
@@ -14,23 +14,13 @@ BackLinkHelper.prototype.init = function init() {
        Currently, a page refresh sets the referer to empty, leading to the back link being hidden
        under our existing logic.
      */
-    const invalidDomain = () => {
-      const ref = this.document.referrer;
-      let isInvalid = false;
-
-      if (ref) {
-        const diffDom = ref.indexOf(this.window.location.host) === -1;
-        if (diffDom) {
-          // Allow PEGA domains
-          const allowList = ['account-np.hmrc.gov.uk', 'account.hmrc.gov.uk'];
-          const allowed = allowList.some((e) => ref.includes(e));
-          isInvalid = !allowed;
-        }
-      } else {
-        isInvalid = true;
-      }
-
-      return isInvalid;
+    const referrerAllowList = ['account-np.hmrc.gov.uk', 'account.hmrc.gov.uk'];
+    const shouldHideBackLink = () => {
+      const referrer = this.document.referrer;
+      if (!referrer) return true
+      const referredFromDifferentDomain = referrer.indexOf(this.window.location.host) === -1;
+      const referrerNotOnAllowList = !referrerAllowList.some((allowListedDomain) => referrer.includes(allowListedDomain));
+      return referredFromDifferentDomain && referrerNotOnAllowList
     };
 
     // hide the backlink if the referrer is on a different domain or the referrer is not set
--- a/src/components/back-link-helper/back-link-helper.js
+++ b/src/components/back-link-helper/back-link-helper.js
@@ -14,19 +14,20 @@ BackLinkHelper.prototype.init = function init() {
        Currently, a page refresh sets the referer to empty, leading to the back link being hidden
        under our existing logic.
      */
-    const referrerNotOnSameDomain = () => {
-      const referer = this.document.referrer;
-      const noReferer = !referer || referer.indexOf(this.window.location.host) === -1;
+    const invalidDomain = () => {
+      const ref = this.document.referrer;
+      const noRef = !ref;
+      const diffDom = ref.indexOf(this.window.location.host) === -1;
 
-      // Allows PEGA domains
-      const allowedDomains = ['maccount-np.hmrc.gov.uk', 'account.hmrc.gov.uk', 'https://catalogue.tax.service.gov.uk/'];
-      const isAllowed = (!noReferer) ? allowedDomains.some((e) => referer.includes(e)) : false;
+      // Allow PEGA domains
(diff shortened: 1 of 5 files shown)


src/components/back-link-helper/back-link-helper.js
```diff
@@ -14,10 +14,17 @@ BackLinkHelper.prototype.init = function init() {
...
     const referrerNotOnSameDomain = () => {
       const referer = this.document.referrer;
-      return !referer || referer.indexOf(this.window.location.host) === -1;
+
+      // Allows PEGA domains
+      const allowedDomains = ['maccount-np.hmrc.gov.uk', 'account.hmrc.gov.uk'];
+      const isAllowed = allowedDomains.some((e) => referer.includes(e));
+
+      const noReferer = !referer;
+      const sameDomain = referer.indexOf(this.window.location.host) === -1;
```
oscarduignan: I think sameDomain here might be misnamed, I think it's checking if it's on a different domain - it looks like the changes have broken the existing tests and we've not added any new tests to cover this change

then I think rather than modifying the referrerNotOnSameDomain it might be clearer to add a new check, maybe "referrerNotOnSameDomain and referrerNotOnAllowList" type deal - because it seems we've changed the meaning of referrerNotOnSameDomain without changing it's name

src/components/back-link-helper/back-link-helper.js
```diff
@@ -14,10 +14,15 @@ BackLinkHelper.prototype.init = function init() {
...
       const referer = this.document.referrer;
-      return !referer || referer.indexOf(this.window.location.host) === -1;
+      const noReferer = !referer || referer.indexOf(this.window.location.host) === -1;
+
+      // Allows PEGA domains
+      const allowedDomains = ['maccount-np.hmrc.gov.uk', 'account.hmrc.gov.uk', 'https://catalogue.tax.service.gov.uk/'];
+      const isAllowed = (!noReferer) ? allowedDomains.some((e) => referer.includes(e)) : false;
+
+      return (isAllowed) ? false : noReferer;
     };
```
oscarduignan: the logic of this is quite hard to read with the multiple conditionals, noreferer now also contains logic to check "no referer or referer is a different domain", what about something like:

```javascript
    const referer = this.document.referrer;

    const referrerNotOnSameDomain = () =>
      (referer.indexOf(this.window.location.host) === -1);
    
    const allowedDomains = ...
    const referrerNotOnAllowList = () =>
      !allowedDomains.some(allowedDomain => referer.includes(allowedDomain));

    if (!referrer || (referrerNotOnSameDomain() && referrerNotOnAllowList())) {
      this.$module.classList.add('hmrc-hidden-backlink');
```

might have some syntactic mistakes, just written out in comment - would still need some automated tests it works

src/components/back-link-helper/back-link-helper.js
```diff
@@ -17,10 +17,10 @@ BackLinkHelper.prototype.init = function init() {
     const invalidDomain = () => {
       const ref = this.document.referrer;
       const noRef = !ref;
-      const diffDom = ref.indexOf(this.window.location.host) === -1;
+      const diffDom = (ref) ? ref.indexOf(this.window.location.host) === -1 : false;
 
       // Allow PEGA domains
-      const allowList = ['maccount-np.hmrc.gov.uk', 'account.hmrc.gov.uk'];
+      const allowList = ['maccount-np.hmrc.gov.uk', 'account.hmrc.gov.uk','https://jira.tools.tax.service.gov.uk/'];
```
oscarduignan: dave just checking, why have we added jira in here? I don't know how private it is but this is an internal url

oscarduignan: I think I saw the catalogue appear before too

goodvibes: Just using for testing logic in console, won't  be in final

src/components/back-link-helper/back-link-helper.js
```diff
@@ -14,14 +14,27 @@ BackLinkHelper.prototype.init = function init() {
...
+          // Allow PEGA domains
+          const allowList = ['account-np.hmrc.gov.uk', 'account.hmrc.gov.uk'];
+          const allowed = allowList.some((e) => ref.includes(e));
+          isInvalid = !allowed;
+        }
+      } else {
+        isInvalid = true;
+      }
+
+      return isInvalid;
```
oscarduignan: we can simplify the logic here with an early return, and I think invalidDomain doesn't tell us much about the intent of what we're checking, how about something like this:
```suggestion
    const referrerAllowList = ['account-np.hmrc.gov.uk', 'account.hmrc.gov.uk'];
    const shouldHideBackLink = () => {
      const referrer = this.document.referrer;
      if (!referrer) return true
      const referredFromDifferentDomain = referrer.indexOf(this.window.location.host) === -1;
      const referrerNotOnAllowList = !referrerAllowList.some((allowListedDomain) => referrer.includes(allowListedDomain));
      return referredFromDifferentDomain && referrerNotOnAllowList
```
I don't think we're saving much by shortening variable names, it's more readable to use the full words - for example, I think "ref" and "diffDom" are a bit ambiguous

edit: I had reversed the logic in my example 🤦‍♂️

oscarduignan: one gap this highlights for me too, is that the current implementation would be so easy to circumvent by putting the current url in a query param on the referring url

oscarduignan: I think it's good that we're removing it in the future

oscarduignan (dismissed):
looks good to me 🙌

hmrc-web-operations:

oscarduignan:
test this please

TimothyFothergill (approved):
lgtm

PR #422, opened 2024-12-10
PLATUI-3359 ensure timeout dialog tests resolve gracefully

TimothyFothergill:
# Purpose of PR
- Ensure tests for timeout dialog resolve gracefully
  - We had an issue where sometimes, the timeout dialog would time out during tests, resulting in a fail for the `should update the audible time remaining every 20 seconds` test
    - This issue was flaky, as sometimes it would get past this test with no issue
    - To do this, I dropped a second from a check, so the method `twentyTimes` is now `nineteenTimes`.
      - This should be safe, as the second between is the second when we would expect the text to update

### DIFF
--- a/lib/browser-tests/puppeteer-helpers.js
+++ b/lib/browser-tests/puppeteer-helpers.js
@@ -41,7 +41,6 @@ export function withHmrcStylesAndScripts(body) {
     <head>
       <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover">
       <link rel="stylesheet" href="/assets/hmrc-frontend-${version}.min.css">
-      <link rel="stylesheet" href="/assets/accessible-autocomplete-${version}.css">
       ${preloadGovukFonts}
     </head>
     <body class="govuk-template__body">
@@ -50,7 +49,6 @@ export function withHmrcStylesAndScripts(body) {
       </script>
         ${body}
         <script src="/assets/hmrc-frontend-${version}.min.js" type="module"></script>
-        <script src="/assets/accessible-autocomplete-${version}.js" type="module"></script>
     </body>
     </html>
   `;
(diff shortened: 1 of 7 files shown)


lib/browser-tests/puppeteer-helpers.js
```diff
@@ -41,6 +41,7 @@ export function withHmrcStylesAndScripts(body) {
     <head>
       <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover">
       <link rel="stylesheet" href="/assets/hmrc-frontend-${version}.min.css">
+      <link rel="stylesheet" href="/assets/accessible-autocomplete-${version}.css">
```
TimothyFothergill: Maybe I don't need to bring these in as part of this PR?

oscarduignan: yup not needed 🙌

lib/browser-tests/puppeteer-helpers.js
```diff
@@ -49,19 +50,27 @@ export function withHmrcStylesAndScripts(body) {
...
+    return Promise.resolve().then(() => req.continue()).catch(() => {});
+  };
+  page.on('request', interceptPageRender);
+  try {
+    await page.goto('http://localhost:3000/', options); // if ever flaky, waitUntil networkidle0 (js loaded)
+    await page.bringToFront();
+  } finally {
+    page.off('request', interceptPageRender);
+    await page.setRequestInterception(false);
+  }
```
oscarduignan: I've no idea if these changes are needed either if the flakeyness was around that sampling of messages 🤷

TimothyFothergill: Agreed - It seems like the only change that was needed was just that. I'll get these removed and update from draft to ready. Noticed Jo's also going to be running it locally, so will wait until she's had a look too 🫡

oscarduignan (approved):

PR #424, opened 2024-12-16
PLATUI-3352: add failing tests for accessible-autocomplete to show gaps that adam's patches fix ahead of patching them

oscarduignan:


src/components/accessible-autocomplete/accessible-autocomplete.js
```diff
@@ -17,6 +17,7 @@ AccessibleAutoComplete.prototype.init = function init() {
       autoselect,
       defaultValue,
       minLength,
+      confirmOnBlur: true,
```
oscarduignan: just added to check, it's the default but I think I've misunderstood something about how this works - because `should select a matching option in the underlying select when input is changed if there is one` isn't working as I would expect

oscarduignan:
finished the initial tests for this I think, I've marked the ones that won't work until we implement the fixes as failing so they could be merged ahead of other one if we want to review the tests for now

lib/jest-helpers.js
```diff
@@ -22,9 +22,9 @@ nunjucks.configure(
  * @param {string} componentName
  * @param {string} params parameters that are used in the component macro
  * @param {any} children any child components or text, pass the children to the macro
- * @returns {function} returns cheerio (jQuery) instance of the macro for easy DOM querying
+ * @returns {string} returns html
  */
-function render(componentName, params, children = false) {
+function renderString(componentName, params, children = false) {
```
oscarduignan, line 27: split this out because I wanted the string, could have possibly gotten the string back out from cheerio though instead, :hmm:

src/components/accessible-autocomplete/accessible-autocomplete.browser.test.js
```diff
@@ -0,0 +1,345 @@
...
+      ],
+    }));
+
+    await expect(page).toFill('#location', 'Lon');
+    await acceptFirstSuggestionFor('#location');
+    expect(await page.$eval('select', (select) => select.value)).toBe('london');
+    await expect(page).toFill('#location', 'South West');
+    await page.$eval('#location', (input) => input.blur());
+    expect(await page.$eval('select', (select) => select.value)).toBe('southwest');
+  });
```
oscarduignan, line 188: this is the bit that I thought was additional behaviour, but actually I think adam's patch will be achieving this, it's just that my approach that I proof of concepted wouldn't be - onConfirm will get "undefined" in this case, so you have to in this case lookup the current value of the autocomplete input instead and use that to mark an option as selected

oscarduignan:
~I'll update changelog tomorrow~ (decided not to do this, think it's snot worth a new release, will mean 1 failed build but that's ok)

can merge these new tests in now - and in the implementation PR just remove the ".failing" modifier from the ones we are implementing

TimothyFothergill (approved):
lgtm

PR #430, opened 2024-12-20
PLATUI-3352: patch accessible autocomplete

oscarduignan:
~also swaps to use playwright for backstopjs~

~because we needed to udpate playwright and the latest version was timing out a lot when using puppeteer~

playwright already swapped in separately


.eslintrc.json
```diff
@@ -27,6 +27,9 @@
   "rules": {
     "import/prefer-default-export": "off"
   },
+  "parserOptions": {
+    "ecmaVersion": "latest"
+  },
   "overrides": [
```
oscarduignan, line 33: I think new stuff will still be transpiled away by babel, but eslint would reject using new (well supported) stuff before this - this lets us use it - would be good to review how we are bundling stuff with babel/rollup/gulp to more closely match govuk-frontend as a default so we aren't reinventing stuff if we consider them a safe default (future stuff)

.gitignore
```diff
@@ -43,4 +43,4 @@ backstop_data/html_report/*
 
 webjar/
 webjar-dist/
-
+core
```
oscarduignan, line 46: when I run backstop this core file gets generated? some artifact of docker

.npmrc
```diff
@@ -1,2 +1 @@
-PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
```
oscarduignan: now using playwright

backstop_data/engine_scripts/puppet/onBefore.js
```diff
@@ -1,3 +1,3 @@
 module.exports = async (page, scenario, vp) => {
-  await require('./loadCookies')(page, scenario);
+  // await require('./loadCookies')(page, scenario);
```
oscarduignan: not sure what this was used for, should be removed

src/components/accessible-autocomplete/accessible-autocomplete.js
oscarduignan, line 1: will split these changes out

oscarduignan:
will finish this off on monday, just pushing because I want to see if playwright runs ok in build or if it fails because it's blocked - if so then we need to get it unblocked / or figure out how to get it from artifactory

worked ok, just had to run `playwright install` first - can ignore the errors about deps missing as we don't use the things those deps are for - the slight "eh" now from this is that it means we are now downloading playwright binaries and puppeteer binaries, and playwright is installing more than we need (edge and firefox)

backstop_data/bitmaps_reference/backstop_default_accessible-autocomplete_with-show-all-values_example_accessibility-autocomplete-with-large-font_0_document_0_phone.png
oscarduignan, line 1: I tried a few small css tweaks to fix this case, but I couldn't get anything that I was happy with that wouldn't possibly break other stuff so have left fixing this out for now - if someone has increased their base font size then the position of the dropdown arrow won't be centered

src/components/accessible-autocomplete/accessible-autocomplete.js
```diff
@@ -6,20 +6,34 @@ function AccessibleAutoComplete($module, window, document) {
...
+      selectElement,
       showAllValues,
       autoselect,
       defaultValue,
       minLength,
+      onConfirm: (chosenOption) => {
+        selectElement.value = '';
+        const chosenOptionOrCurrentValue = (typeof chosenOption !== 'undefined')
+          ? chosenOption
+          : document.getElementById(autocompleteId)?.value;
```
oscarduignan, line 27: just to double check, I looked at the compiled output and `?.` is not passed through, so won't be any compatibility issues by using it here

PatrykRudnicki (dismissed):
LGTM

TimothyFothergill (approved):
lgtm

PR #431, opened 2024-12-23
PLATUI-3352 update playwright

oscarduignan:
because it seems to work a bit more reliably (where puppeteer was timing out)


package.json
```diff
@@ -46,14 +46,14 @@
     "govuk-frontend": "^5.7.1"
   },
   "devDependencies": {
-    "sass": "1.77.6",
     "@babel/preset-env": "^7.21.5",
+    "@playwright/browser-chromium": "^1.49.1",
```
oscarduignan: this dependency means that playwright chromium browser will be downloaded on install

package.json
```diff
@@ -102,6 +102,7 @@
     "rollup-plugin-babel": "^4.4.0",
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-node-resolve": "^5.2.0",
+    "sass": "1.77.6",
```
oscarduignan, line 106: npm doing npm things, it reshuffled this position for some reason

package.json
```diff
@@ -46,14 +46,14 @@
...
-    "sass": "1.77.6",
     "@babel/preset-env": "^7.21.5",
+    "@playwright/browser-chromium": "^1.49.1",
     "adm-zip": "^0.5.5",
     "adr-log": "^2.2.0",
     "ajv": "^8.12.0",
     "autoprefixer": "^9.6.1",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
-    "backstopjs": "^6.1.4",
+    "backstopjs": "6.3.25",
```
oscarduignan, line 55: updating to this without switching to playwright and now being on an arm mac was causing test instability

oscarduignan:
hmm, hit a few speed bumps in investigating this, I have updated the playwright version ahead of the one used by backstop, which caused a load of reference images to need updating locally - I'm not sure if this will pass the backstop tests on jenkins, if it fails then I'll downgrade the playwright version to the one used by backstop and try that

edit: of course, made a silly mistake here, I was testing that playwright downloads stuff in the way that I expect, and using "env CI=true" to check that it used local playwright that was downloaded rather than docker one, I ran backstop tests with CI=true still and that took the screenshots via local browser which will never match up to jenkins ones (where the screenshots taken in docker will)

oscarduignan:
seems like without installing an explicit version of playwright-core and playwright-chromium and overriding playwright to that version of playwright-core, sometimes running the backstop tests can try to use the wrong version of playwright

.npmrc
```diff
@@ -1,2 +1 @@
-PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
```
oscarduignan, line 1: also removed from build jobs where it was also defined

package.json
```diff
@@ -114,5 +116,8 @@
     "ie 8",
     "ie 9",
     "iOS 9"
-  ]
+  ],
+  "overrides": {
+    "playwright": "$playwright-core"
```
oscarduignan, line 121: otherwise sometimes it can try to use a playwright browser we didn't download on install

src/components/timeout-dialog/timeout-dialog.browser.test.js
```diff
@@ -252,9 +252,10 @@ describe('/components/timeout-dialog', () => {
         '2 seconds',
         '1 second',
       ]);
-
-      await clockTickSeconds(page, 1);
-      await page.waitForNavigation({ timeout: 1000 });
+      // sometimes we have already reached timeout page
+      // so we can discard exceptions after this point
+      await clockTickSeconds(page, 1).catch(() => {});
+      await page.waitForNavigation({ timeout: 1000 }).catch(() => {});
```
oscarduignan, line 258: tested this locally by adding in another clockTickSeconds before it to recreate the bug 100% of the time

seems like sometimes we can trigger the timeout when we are capturing the final "time remaining" message, not sure exactly what it is with the clock causing that, maybe sometimes it can be init after a few seconds of the page loading and it starts at non-zero? rather than try to isolate that which might be really hard, we can just ignore exceptions if the redirect on timeout already happened

oscarduignan: no other stuff I think where we are testing stuff with such a narrow margin

tasks/gulp/backstop-config.js
```diff
@@ -78,7 +78,7 @@ module.exports = ({ host, port, components }) => ({
     ci_report: 'backstop_data/ci_report',
   },
   report: ['browser'],
-  engine: 'puppeteer',
+  engine: 'playwright',
```
oscarduignan, line 81: the apis we are using are compatible, no idea if there are some we have coded but aren't using that won't be

TimothyFothergill (approved):
lgtm

Branch tested locally on Ubuntu 24.04 and working fine here too

PR #432, opened 2025-01-08
PLATUI-3352: document a possible bug for fixing in the future with autocomplete

oscarduignan:
no package.json version bump because it's just a test change, we've asked DIAS about this but since it wasn't actually part of adam's patch, it seems like something we shouldn't rush to fix in case we make a mistake and impact screen reader users

### DIFF
--- a/src/components/accessible-autocomplete/accessible-autocomplete.browser.test.js
+++ b/src/components/accessible-autocomplete/accessible-autocomplete.browser.test.js
@@ -82,7 +82,7 @@ describe('Patched accessible autocomplete', () => {
   // created, we're not sure how this should actually behave, so we're
   // going to check with the DIAS team before we implement a fix and for
   // now just document that this happens.
-  it.failing('should announce still announce the hint and error message linked to the underlying select after interaction with field', async () => {
+  it.failing('should still announce the hint and error message linked to the underlying select after interaction with field', async () => {
     await render(page, withGovukSelect({
       id: 'location',
       name: 'location',
--- a/src/components/accessible-autocomplete/accessible-autocomplete.browser.test.js
+++ b/src/components/accessible-autocomplete/accessible-autocomplete.browser.test.js
@@ -78,6 +78,45 @@ describe('Patched accessible autocomplete', () => {
     expect(ariaDescribedBy).toBe('location-hint location-error location__assistiveHint');
   });
 
+  // This wasn't something covered by the original polyfill that adam
+  // created, we're not sure how this should actually behave, so we're
+  // going to check with the DIAS team before we implement a fix and for
+  // now just document that this happens.
+  it.failing('should announce still announce the hint and error message linked to the underlying select after interaction with field', async () => {
+    await render(page, withGovukSelect({
+      id: 'location',
+      name: 'location',
+      attributes: {
+        'data-module': 'hmrc-accessible-autocomplete',
+      },
+      label: {
+        text: 'Choose location',
+      },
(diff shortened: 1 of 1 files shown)


src/components/accessible-autocomplete/accessible-autocomplete.browser.test.js
```diff
@@ -78,6 +78,45 @@ describe('Patched accessible autocomplete', () => {
     expect(ariaDescribedBy).toBe('location-hint location-error location__assistiveHint');
   });
 
+  // This wasn't something covered by the original polyfill that adam
+  // created, we're not sure how this should actually behave, so we're
+  // going to check with the DIAS team before we implement a fix and for
+  // now just document that this happens.
+  it.failing('should announce still announce the hint and error message linked to the underlying select after interaction with field', async () => {
```
oscarduignan:
```suggestion
  it.failing('should still announce the hint and error message linked to the underlying select after interaction with field', async () => {
```

src/components/accessible-autocomplete/accessible-autocomplete.browser.test.js
```diff
@@ -78,6 +78,45 @@ describe('Patched accessible autocomplete', () => {
...
+    }));
+
+    const element = await page.$('#location');
+    const tagName = await element.evaluate((el) => el.tagName.toLowerCase());
+    await expect(page).toFill('#location', 'London');
+    await page.$eval('#location', (input) => input.blur());
+    const ariaDescribedBy = await element.evaluate((el) => el.getAttribute('aria-describedby'));
+
+    expect(tagName).not.toBe('select'); // or select element was not enhanced to be an autocomplete component
+    expect(ariaDescribedBy).toBe('location-hint location-error location__assistiveHint');
```
oscarduignan, line 117: what we actually get here is "null" I would have thought that it would display "location__assistiveHint" which is a message that describes how you can interact with the field, but it seems like that isn't always present

aha, [from the docs](https://github.com/alphagov/accessible-autocomplete/blob/96bf629a2c02b98bbd924dbc575692730e5789ee/README.md#tassistivehint-default---when-autocomplete-results-are-available-use-up-and-down-arrows-to-review-and-enter-to-select--touch-device-users-explore-by-touch-or-with-swipe-gestures)
> This text is intended as an initial instruction to the assistive tech user. The `aria-describedby` attribute is automatically removed once user input is detected, in order to reduce screen reader verbosity.

That's what we're hitting, feels like maybe the hint and error should still be associated with it - but maybe it's good enough for it to be announced once on initial interaction

oscarduignan: I will update the comment slightly to incorporate a link to this

oscarduignan, line 117:
```suggestion
    expect(ariaDescribedBy).toBe('location-hint location-error location__assistiveHint');
    // The value will currently be null, because the autocomplete removes the
    // link to location__assistiveHint to reduce screen reader verbosity.
    // Where the assistiveHint explains how you can interact with the input.
```

oscarduignan: updated to add an explanation of the current value and why it's null

oscarduignan:
we got quick confirmation by DIAS so we don't need this PR anymore, it's superseded now by https://github.com/hmrc/hmrc-frontend/pull/434

PR #438, opened 2025-01-23
NOJIRA refactor request interception in browser tests to reduce flakiness

oscarduignan:
not sure I've totally understood all the nuance, but this seems to do the trick for me - will see if I can get other devs on team to run this branch locally a load of times and check they don't run into any flakiness

### DIFF
--- a/lib/browser-tests/puppeteer-helpers.js
+++ b/lib/browser-tests/puppeteer-helpers.js
@@ -1,5 +1,6 @@
 import path from 'path';
 import fs from 'fs';
+import crypto from 'crypto';
 import { version } from '../../package.json';
 
 export function delay(ms) {
@@ -57,13 +58,25 @@ export function withHmrcStylesAndScripts(body) {
 }
 
 export async function render(page, body, options) {
+  const temporaryPageUrl = `http://localhost:3000/${crypto.randomUUID()}`;
   await page.setRequestInterception(true);
-  page.once('request', (req) => {
-    req.respond({ contentType: 'text/html', body });
-    page.setRequestInterception(false);
-  });
-  await page.goto('http://localhost:3000/', options); // if ever flaky, waitUntil networkidle0 (js loaded)
+  const serveTemporaryPage = async (req) => {
+    if (req.isInterceptResolutionHandled()) return;
+
+    if (req.url() === temporaryPageUrl) {
+      await req.respond({ contentType: 'text/html', body });
+    } else {
+      // we need to fulfill requests for css and fonts
+      // or our tests will hang and fail after 30 seconds
+      // with a navigation timeout.
+      await req.continue();
+    }
+  };
+  page.on('request', serveTemporaryPage);
+  await page.goto(temporaryPageUrl, options);
   await page.evaluateHandle('document.fonts.ready');
   await page.bringToFront();
+  page.off('request', serveTemporaryPage);
+  await page.setRequestInterception(false);
   return page;
 }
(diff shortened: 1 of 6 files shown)


lib/browser-tests/puppeteer-helpers.js
```diff
@@ -57,13 +58,25 @@ export function withHmrcStylesAndScripts(body) {
 }
 
 export async function render(page, body, options) {
+  const temporaryPageUrl = `http://localhost:3000/${crypto.randomUUID()}`;
   await page.setRequestInterception(true);
-  page.once('request', (req) => {
-    req.respond({ contentType: 'text/html', body });
-    page.setRequestInterception(false);
-  });
```
oscarduignan, line 64: from the puppeteer docs

> Once request interception is enabled, every request will stall unless it's continued, responded or aborted.

so it seems like it was possible for us to respond, and the browser to load the page and request static assets before setRequestInterception(false) kicked in

but I wasn't able to pin it down exactly, I tried adding delays and awaits and reordering stuff

I was able to trigger the 30s navigation timeout after reasonably often

with that idea of what was going wrong in mind, I reordered stuff so that we explicitly continue any other requests that are triggered by rendering the page in the browser - and if you comment out line 72 then for me that reliably triggered the 30s timeout

after this refactoring I haven't been able to trigger the 30s navigation timeout from multiple runs

but it would be useful for others to try this out

one other thing that never seemed to trigger it, was using --runInBand (so that tests are only run one at a time)

if we still get flakiness in jenkins, then using that when building there might be a solution - because we probably don't mind increasing the test run time in that context

take away for me - request interception seems to be quite complicated, nuanced, and difficult to debug

there's other levels to it too - cooperative mode 🤯

src/components/accessible-autocomplete/accessible-autocomplete.browser.test.js
```diff
@@ -25,22 +25,24 @@ function acceptFirstSuggestionFor(autocompleteSelector) {
...
-  });
+    await request.continue();
+  };
+  page.on('request', capturePostData);
   return {
-    postedFormData,
+    postedFormData: postedFormData.finally(async () => {
+      page.off('request', capturePostData);
+      await page.setRequestInterception(false);
+    }),
```
oscarduignan, line 45: refactored to remove the handler after so that we can never get "request already handled" or "request interception not turned on" errors

src/components/timeout-dialog/timeout-dialog.browser.test.js
```diff
@@ -193,12 +193,13 @@ describe('/components/timeout-dialog', () => {
     `);
     await clockTickSeconds(page, 899);
     await page.setRequestInterception(true);
-    await page.once('request', (req) => {
-      setTimeout(() => req.respond({
+    page.once('request', async (req) => {
```
oscarduignan, line 196: page.once doesn't return a promise so no need to await it

src/components/timeout-dialog/timeout-dialog.browser.test.js
```diff
@@ -193,12 +193,13 @@ describe('/components/timeout-dialog', () => {
...
+      await delay(2000);
+      await req.respond({
         contentType: 'text/plain',
         body: 'simulate slow response to signing out',
-      }), 2000);
-      page.setRequestInterception(false);
+      });
+      await page.setRequestInterception(false);
```
oscarduignan, line 202: good to await here, after tests we reset the page, if you don't await this then I think you could run into case where this happens after page is reset and gives an error (different possibly bit of flakiness)

JoPintoPaul:
Can confirm that multiple runs of these tests pass when running on:
* MacBook Air 2020 with Intel Core i7
* macOS Ventura 13.5

TimothyFothergill (approved):
lgtm

Ran the branch locally - lgtm!

PR #443, opened 2025-01-31
PLATUI-3473: expand aria-live region in timeout-dialog

### LINKED ISSUE #387: Timeout dialog doesn't read out heading in VoiceOver
I'm doing some quick ad-hoc testing of the timeout dialog in Voiceover on Safari on Mac and iOS.

In both, when the dialog is shown, the paragraph gets read out, but not the heading. My expectation is that the heading should get read out too - at least on first load.
Screenshot 2024-07-08 at 15 59 37

Possibly relevant, I note the paragraph gets `aria-live="assertive"` but the h1 does not.

oscarduignan:
# Ensure screen readers don't skip heading in timeout dialog when it opens

**Bug fix**

Took a while to get here, and we ended up with what we tried first I think but for some reason the first time it didn't work but now it does, not sure what was different

All we needed to do was expand aria-live region, and only elements that are updated are re-announced, not the whole thing

Subject to needing to test this with JAWS and NVDA

So far tested with VoiceOver on IOS and mac

Fixes #387 

## Checklist

### DIFF
--- a/src/components/timeout-dialog/timeout-dialog.js
+++ b/src/components/timeout-dialog/timeout-dialog.js
@@ -142,7 +142,7 @@ function TimeoutDialog($module, $sessionActivityService) {
   };
 
   const setupDialog = (signoutTime) => {
-    const $element = utils.generateDomElementFromString('<div>');
+    const $element = utils.generateDomElementFromString('<div aria-live="assertive">');
 
     if (settings.title) {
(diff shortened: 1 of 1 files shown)


oscarduignan:
think in the end we decided to do nothing about this - can't remember reason but will be somewhere in tickets we worked on this on in jira

PR #444, opened 2025-02-06
Plat UI 3484 handle deprecation warnings for sass compiler

TimothyFothergill:
# Handle deprecation warnings for SASS compiler

**New feature**

- Introduces suppression to the sass deprecations
- Small description update in MAINTAINERS to explain when we might want to suppress rather than fix
- Update `gulp-sass` to latest (`6.0.0`)
  - This includes a small change of `includePaths` to `loadPaths` for gulp settings 

## Checklist

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -81,6 +81,10 @@ We use the following workflow for patch releases:
         commit tag:"6.0.0" id:"latest"
 ```
 
+## Consider if a deprecation warning needs a change request
+
+Sometimes we may get deprecation warnings, but we might not always want to resolve this with a change request. This is because these changes can be breaking changes for the consumers of our library. Instead, you can suppress the deprecation by adding the name of the deprecation to the `silenceDeprecations` list and documenting the suppression.
+
 ## Keeping a record of decisions
 
 We are using MADRs to record significant decisions in this service. To find out more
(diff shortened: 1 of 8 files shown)


PatrykRudnicki (dismissed):
LGTM

MAINTAINERS.md
```diff
@@ -81,6 +81,10 @@ We use the following workflow for patch releases:
         commit tag:"6.0.0" id:"latest"
 ```
 
+## Consider if a deprecation warning needs a change request
+
+Sometimes we may get deprecation warnings, but we might not always want to resolve this with a change request. This is because these changes can be breaking changes for the consumers of our library. Instead, you can suppress the deprecation by adding the name of the deprecation to the `silenceDeprecations` list and documenting the suppression.
```
oscarduignan:
```suggestion
Sometimes we may get deprecation warnings, but we might not always want to resolve this with a change request. This is because these changes can be breaking changes for the consumers of our library who reuse our uncompiled sass. The largest group of which are prototypes. Instead, you can suppress the deprecation by adding the name of the deprecation to the `silenceDeprecations` list and documenting the suppression.
```

oscarduignan: in my head I'm thinking it would be easy to think that because we compile our sass to css, that deprecations won't be breaking. And the integration of our styles with prototypes might be non-obvious for future people joining.

PatrykRudnicki (approved):

oscarduignan (approved):

PR #450, opened 2025-03-20
PLATUI-2019 allow trailing/leading whitespace in accessible-autocomplete inputs

TimothyFothergill:
# Allow trailing/leading whitespace in accessible-autocomplete inputs

**Bug fix** (delete as appropriate)

- Added a source function to trim whitespace on queries

Fixes # https://github.com/hmrc/accessibility-audits-external/issues/2193 

## Checklist

### DIFF
--- a/src/components/accessible-autocomplete/accessible-autocomplete.js
+++ b/src/components/accessible-autocomplete/accessible-autocomplete.js
@@ -6,6 +6,12 @@ function AccessibleAutoComplete($module, window, document) {
 
 AccessibleAutoComplete.prototype.init = function init() {
   if (this.$module) {
+    const trimQuery = (values) => (query, syncResults) => {
+      const matches = values.filter((r) => r.toLowerCase()
+        .indexOf(query.toLowerCase().trim()) !== -1);
+      syncResults(matches.filter((item) => (item)));
+    };
+
     const selectElement = this.$module;
     const selectOptions = Array.from(selectElement.options);
     const autocompleteId = selectElement.id;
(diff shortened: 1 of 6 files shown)


src/components/accessible-autocomplete/accessible-autocomplete.js
```diff
@@ -6,6 +6,12 @@ function AccessibleAutoComplete($module, window, document) {
 
 AccessibleAutoComplete.prototype.init = function init() {
   if (this.$module) {
+    const trimQuery = (values) => (query, syncResults) => {
+      const matches = values.filter((r) => r.toLowerCase()
+        .indexOf(query.toLowerCase().trim()) !== -1);
+      syncResults(matches.filter((item) => (item)));
```
oscarduignan, line 12: I'm not sure if there might be something in here slightly different in terms of handling when there's an option with no value / empty value / falsey value

because enhanceSelectElement builds the source like this https://github.com/alphagov/accessible-autocomplete/blob/96bf629a2c02b98bbd924dbc575692730e5789ee/src/wrapper.js#L24-L25

which looks a bit different, but I haven't properly looked through it so I might just be saying something misdirecting here - I will take more of a look after lunch if that's ok or feel free to correct me before then

oscarduignan:
I think we should add some test cases to show the change in behaviour

src/components/accessible-autocomplete/accessible-autocomplete.js
TimothyFothergill: In the syncResults I'm doing another matches.filter. This ensures if there's empty/falsey values, it drops them. The POC was failing as when it runs and you have with-show-all-values, it added the empty value in.

PR #452, opened 2025-03-28
PLATUI-2019 trim accessible-autocomplete trailing/leading whitespace

TimothyFothergill:
# Allow trailing/leading whitespace in accessible-autocomplete inputs

**Bug fix** (delete as appropriate)

- Added a source function to trim whitespace on queries

Fixes # https://github.com/hmrc/accessibility-audits-external/issues/2193 

## Checklist

### DIFF
--- a/src/components/accessible-autocomplete/accessible-autocomplete.js
+++ b/src/components/accessible-autocomplete/accessible-autocomplete.js
@@ -9,7 +9,7 @@ AccessibleAutoComplete.prototype.init = function init() {
     const trimQuery = (values) => (query, syncResults) => {
       const matches = values.filter((r) => r.toLowerCase()
         .indexOf(query.toLowerCase().trim()) !== -1);
-      syncResults(matches.filter((item) => (item)));
+      syncResults(matches);
     };
     const selectElement = this.$module;
     const selectOptions = Array.from(selectElement.options);
@@ -25,7 +25,11 @@ AccessibleAutoComplete.prototype.init = function init() {
       autoselect,
       defaultValue,
       minLength,
-      source: trimQuery(Array.from(this.$module.options).map((a) => a.textContent)),
+      // we don't yet support preserveNullOptions,
+      // but if we start then it needs to override this filtering
+      // https://github.com/alphagov/accessible-autocomplete/blob/main/src/wrapper.js#L24
+      source: trimQuery(Array.from(this.$module.options).filter((a) => a.value)
+        .map((a) => a.textContent)),
       onConfirm: (chosenOption) => {
         selectElement.value = '';
         const chosenOptionOrCurrentValue = (typeof chosenOption !== 'undefined')
--- a/src/components/accessible-autocomplete/accessible-autocomplete.js
+++ b/src/components/accessible-autocomplete/accessible-autocomplete.js
@@ -6,6 +6,11 @@ function AccessibleAutoComplete($module, window, document) {
 
 AccessibleAutoComplete.prototype.init = function init() {
   if (this.$module) {
+    const trimQuery = (values) => (query, syncResults) => {
+      const matches = values.filter((r) => r.toLowerCase()
+        .indexOf(query.toLowerCase().trim()) !== -1);
+      syncResults(matches.filter((item) => (item)));
+    };
     const selectElement = this.$module;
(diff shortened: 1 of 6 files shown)


src/components/accessible-autocomplete/accessible-autocomplete.js
```diff
@@ -20,6 +25,7 @@ AccessibleAutoComplete.prototype.init = function init() {
       autoselect,
       defaultValue,
       minLength,
+      source: trimQuery(Array.from(this.$module.options).map((a) => a.textContent)),
```
oscarduignan:
```suggestion
      source: trimQuery(Array.from(this.$module.options).filter((a) => {
        // we don't yet support preserveNullOptions, but if we start then it needs to override this filtering
        // https://github.com/alphagov/accessible-autocomplete/blob/main/src/wrapper.js#L24 
        return a.value
      }).map((a) => a.textContent)),
```

src/components/accessible-autocomplete/accessible-autocomplete.js
```diff
@@ -6,6 +6,11 @@ function AccessibleAutoComplete($module, window, document) {
 
 AccessibleAutoComplete.prototype.init = function init() {
   if (this.$module) {
+    const trimQuery = (values) => (query, syncResults) => {
+      const matches = values.filter((r) => r.toLowerCase()
+        .indexOf(query.toLowerCase().trim()) !== -1);
+      syncResults(matches.filter((item) => (item)));
```
oscarduignan:
```suggestion
      syncResults(matches);
```

src/components/accessible-autocomplete/accessible-autocomplete.js
oscarduignan, line 1: @TimothyFothergill apologies, I looked at this and thought "how is it working with no empty value filtering?" and only then spotted that some of our examples, which I guess we mostly use for VRT so they aren't always totally correct, were missing a label on their empty values, when you add a label (which services will have) and run the tests then they fail, because the empty values are still there because the filter on matches is happening on the visible text

so we can drop the matches filter I think, and we need to add a filter on the options that we use to populate the source list

example test failure with labels added, so we should also add in the labels as part of this to make sure we have coverage https://github.com/hmrc/hmrc-frontend/pull/453/files

src/components/accessible-autocomplete/accessible-autocomplete.js
oscarduignan: I haven't actually run this, just edited inline in github, so might have got syntax wrong

oscarduignan: also noticed that we have
```
const selectOptions = Array.from(selectElement.options);
```
so feel free to refactor more if you want! Like this maybe (though not totally sure if we need empty options in selectOptions so worth thinking through that doesn't break anything! sort of seems like you could type "Select country" at the moment and choose the empty option, even if it wasn't in the visible option list)
```
const selectOptions = Array.from(selectElement.options).filter(option => option.value);
// ...
source: (query, syncResults) => {
      const matches = selectOptions.filter((r) => r.toLowerCase()
        .indexOf(query.toLowerCase().trim()) !== -1);
      syncResults(matches);
},
```
also looking, this bit down below in onConfirm - this might be a future thing to investigate
```
        const chosenOptionOrCurrentValue = (typeof chosenOption !== 'undefined')
          ? chosenOption
          : document.getElementById(autocompleteId)?.value;
        const selectedOption = [].filter.call(
          selectOptions,
          (option) => (option.textContent || option.innerText) === chosenOptionOrCurrentValue,
        )[0];
```
where we do the filter in "selectedOption" we could also maybe do a lowercase and trim on both sides of the compare too, though maybe there is a case for only automatically matching when it's typed exactly the same (including case and whitespace on ends) 🤷‍♂️ seems bad ux, but bigger scope than how we show options for choosing - since that's impacting what gets selected - where as with options someone has to make a choice still

src/components/accessible-autocomplete/accessible-autocomplete.js
TimothyFothergill: Got it, makes sense! Made those changes now 😄

oscarduignan (approved):

PR #457, opened 2025-05-01
PLATUI-3689 timeout dialog accessibility fixes

TimothyFothergill:
# timeout-dialog accessibility fixes

**Bug fix**

- When on a small screen and zoomed in, these changes allows users to scroll through the component
- Also updated an issue where an errant full stop was appearing on new lines
- Also removed the no-scroll attribute on the html class as this wasn't achieving anything anymore
- Added Visual Regression Test for zoomed in state on mobile

## Checklist


oscarduignan:
could we add a test case where we zoom on mobile to try and recreate the bug and show it's fixed - it looks like you might be able to do it like this https://stackoverflow.com/a/73596252
```
page.evaluate("document.body.style.zoom=2.0")
```

screenshot where I tested (though didn't set the viewport to exact same dimensions)

CleanShot 2025-05-01 at 14 46 11@2x

oscarduignan:
you can do it by adding an allowed action in here https://github.com/hmrc/hmrc-frontend/blob/main/backstop_data/engine_scripts/puppet/beforeTakingScreenshotHelper.js

like "zoom(amount)" and then add a new example for timeout-dialog and add the step in like we do in here https://github.com/hmrc/hmrc-frontend/blob/bd3b62c2f13ec9c30a6e95945b6cf3afc5e79d39/src/components/account-header/account-header.yaml#L165

to have something like:
```
  visualRegressionTests:
    alternateStates:
      zoomed:
        beforeTakingScreenshot:
          - zoom: "200%"
```

src/components/timeout-dialog/_timeout-dialog.scss
```diff
@@ -16,27 +16,24 @@
   -moz-opacity: 0.8;
 }
 
-html {
-  &.no-scroll {
-    overflow: hidden;
-  }
-}
-
```
oscarduignan, line 24: normally - with styles like this which are global, I'd say leave it in _just in case_ - but in this case it's not seemingly ever added to anything (other than one prototype in github org) https://github.com/search?q=org%3Ahmrc+no-scroll&type=code&p=1 

I know we have other orgs we don't know about using our stuff, but in this case it does seem really unused / legacy thing

src/components/timeout-dialog/timeout-dialog.js
```diff
@@ -212,17 +211,17 @@ function TimeoutDialog($module, $sessionActivityService) {
...
-        visibleMessage = `${minutes} ${settings.properties[minutes === 1 ? 'minute' : 'minutes']}`;
+        visibleMessage = `${minutes} ${settings.properties[minutes === 1 ? 'minute' : 'minutes']}.`;
       }
       return visibleMessage;
     };
 
     const getAudibleHumanText = (counter) => {
       const humanText = getHumanText(roundSecondsUp(counter));
-      const messageParts = [settings.message, ' ', humanText, '.'];
+      const messageParts = [settings.message, ' ', humanText];
```
oscarduignan, line 224: @TimothyFothergill what was the reasoning for removing it from the audible? without a period sometimes screen readers won't pause in between bits of text I think

oscarduignan: (not sure if it applies here at all, just double checking reason for changes that I wasn't sure of source of)

TimothyFothergill: Ah I remember now:
humanText now contains the . - When it does getHumanText, it includes the .

TimothyFothergill: https://github.com/hmrc/hmrc-frontend/blob/9bacc29f84412326fc237c4cf9a9682f3c397f26/src/components/timeout-dialog/timeout-dialog.js#L211-L219

oscarduignan: aha! so it's in both still but added in a different way :legend: thank you Tim!

PatrykRudnicki (approved):
LGTM

PR #458, opened 2025-05-19
PLATUI-3745: bump version of govuk-frontend

oscarduignan:
and added a way to flag rebrand in our review app by query param

when viewing an example, add `?rebrand=true` to the url to see that example rebranded, for example

start the review app:
```
fnm install
npm install
npm start
```
then open http://localhost:3000/components/account-header/default/preview?rebrand=true


oscarduignan:
CleanShot 2025-05-19 at 16 34 40@2x

enabled the rebrand, and looks like some of our components don't quite pass the vibe check with no modifications

**gov.uk header fork:**
- the extra header links we add have changed position (presume same will be case for sign out link but haven't checked)
- the contrast of the grey used before for the active item in language selector doesn't pass required contrast ratio for AA accessibility
- the divider in the headers language selector needs updating to something more suitable
- font weight of the extra header links also seems to have reduced, is this ok or does it need changing?

CleanShot 2025-05-19 at 16 38 38@2x

oscarduignan:
my naive approach for now to do tomorrow will be to look at how the styles have been changed for the header links and update our links to match those 🙏

src/components/header/_header.scss
oscarduignan, line 1: all the css in here is scoped under the rebrand

opted for using the class rather than the helper, in case someone installs this in a prototype which hasn't updated to govuk-frontend version with rebrand yet

tasks/gulp/backstop-config.js
```diff
@@ -83,7 +86,7 @@ module.exports = ({ host, port, components }) => ({
     args: ['--no-sandbox'],
   },
   // if running locally, and either stage hangs, try reducing these limits
-  asyncCaptureLimit: 5,
+  asyncCaptureLimit: 3,
```
oscarduignan, line 89: this is reduced because the number of scenarios has doubled in the presence of rebrand for a bit - and having it at 5 was causing some checks to fail for me

tasks/gulp/backstop-config.js
```diff
@@ -69,7 +69,10 @@ module.exports = ({ host, port, components }) => ({
   ],
   onBeforeScript: 'puppet/onBefore.js',
   onReadyScript: 'puppet/onReady.js',
-  scenarios: buildScenarioList(host, port, components),
+  scenarios: buildScenarioList(host, port, components).flatMap((scenario) => [
+    // this means we have twice as many VRT which might increase test time a lot...
+    scenario, { ...scenario, label: `${scenario.label} (using rebrand)`, url: `${scenario.url}?rebrand=true` },
```
oscarduignan, line 74: in the future we might want to reduce this, but for this initial intro of the rebrand it's useful to have a snapshot of all the components with rebrand enabled

src/components/header/_header.scss
```diff
@@ -30,37 +43,94 @@ $govuk-header-link-underline-thickness: 3px;
...
+        color: #ffffff50; // is this acceptable? mimics the existing light grey...
+      }
     }
   }
 }
 
+.govuk-template--rebranded .hmrc-header.hmrc-header--with-additional-navigation .govuk-header__navigation + .hmrc-language-select .govuk-list {
+  padding-top: 0;
+}
+
```
oscarduignan, line 133: this appears to only be used (and maybe only as a potentially configurable option, so may not actually be used in practice) in gforms

there are issues with the current language select implementation in the header in terms of accessibility

teams should not use it, but somewhat we're saying "defer because service nav being introduced to header will clear this up"

TimothyFothergill (approved):
lgtm

PR #460, opened 2025-06-02
PLATUI-3774 bump to govuk-frontend 5.10.2

oscarduignan:
# bump to govuk-frontend 5.10.2

**New feature**

[bump to govuk-frontend 5.10.2 which adjusts padding slightly of govuk-header
](https://github.com/alphagov/govuk-frontend/pull/5968)

It seems like they have changed it so that the padding is not always there, but only added on mobile when there is some navigation items - which is why one of our VRT tests which is mobile but without a menu has changed, the text of our example and screen width are enough that without that extra padding it's able to display header and service name all on one line

I think we must not have a rebranded example impacted maybe - so that's why no other VRT tests are updated, just checked our examples and I can see that our header fork is getting the updates correctly (new class added by js and padding added to logo using it at mobile breakpoint)

CleanShot 2025-06-02 at 10 58 16@2x

## Checklist


PatrykRudnicki (approved):
LGTM

PR #465, opened 2025-06-19
Plat UI 3790 service navigation passable to hmrc-header and new language select component to be used in service navigation

oscarduignan:
# Add service navigation to hmrc header and implement the new language select component for it

**New feature**

You can now pass service navigation to hmrc-header and have it placed in the correct spot in relation to the other banners - following the updated guidance for it's use in hmrc services published on design.tax.service.gov.uk

This include the introduction of a new language select component for use in the service navigation which will replace the language select located in the body of the page

This new language select needs to be passed into the end slot (not navigation end slot) of the service navigation, we've opted for now to not repurpose the languageToggle option of hmrc-header to render the language select into the service navigation so for now it would need to be placed there yourself, for example:

```nunjucks
{% extends "layout.njk" %}

{% from "hmrc/components/header/macro.njk" import hmrcHeader %}
{% from "hmrc/components/service-navigation-language-select/macro.njk" import hmrcServiceNavigationLanguageSelect %}

{% block header %}
  {{ hmrcHeader({
    language: "en",
    serviceNavigation: {
      serviceName: "Example service",
      classes: 'hmrc-service-navigation--with-language-select',
      slots: {
        end: hmrcServiceNavigationLanguageSelect({
          language: "en"
        })
      }
    }
  }) }}
{% endblock %}
```

when we add this into our scala library play-frontend-hmrc, we will be able to hide the need to know where the language switcher should go in the service navigation params from devs

this pr also updates position of the hmrc banner to fit the guidance from gov.uk design system team in the ["navigate a service" pattern](https://design-system.service.gov.uk/patterns/navigate-a-service/) (which is distinct/wider than service navigation component)

## Checklist


backstop_data/bitmaps_reference/backstop_default_service-navigation-language-select_default_example_using_rebrand_0_document_0_phone.png
oscarduignan, line 1: Hey @timsb as part of implementing the new language select component we've noticed that on mobile where the text is smaller when using rebrand it looks like the height of the language select is increasing the height of the service nav a bit- what do you think should change to fix?

src/components/service-navigation-language-select/_service-navigation-language-select.scss
```diff
@@ -0,0 +1,69 @@
...
+
+.hmrc-service-navigation-language-select .govuk-service-navigation__item {
+  margin-bottom: govuk-spacing(3);
+
+  @include govuk-responsive-margin(6, $direction: left);
+
+  @include govuk-media-query($from: tablet) {
+    margin-bottom: 0;
+  }
+}
```
oscarduignan: Hey @timsb we just used the styles from the prototype for now, do we want to use these to start with - and adjust them later when govuk team makes changes or do we want to tweak them more before we release this version?

app/app.js
```diff
@@ -121,15 +121,11 @@ module.exports = (options) => {
...
+    res.locals.componentView = !fs.existsSync(path.join(configPaths.components, componentDirectory, 'macro.njk'))
+      ? ''
+      : env.renderString(`
 {% from '${componentDirectory}/macro.njk' import ${macroName} %}
-{{ ${macroName}(${macroParameters}) }}`,
-      );
-    } catch (err) {
-      res.locals.componentView = null;
-    }
+{{ ${macroName}(${macroParameters}) }}`);
```
oscarduignan, line 128: if there was an error in the nunjucks template then it would swallow it and render a blank page, it also wouldn't log any error- first we tried logging the error then we realised that the reason it was doing that was because some components don't have macros because they are just being used to test stuff or only have js or something like that - we can account for those we think by just checking for the existence of a macro file and in those cases there won't be any rendered component output. Rendering will just be handled by the layoutTemplate

benefit of this is that when there is an error in the template now, it will be displayed in the browser 🙏  so easier to debug

oscarduignan:
haven't rerun the tests since making a few changes, so we will see if the tweaks to the review app to improve err handling blow up any other corner cases 😅 🤞

src/components/service-navigation-language-select/template.njk
```diff
@@ -0,0 +1,41 @@
...
+    <ul class="hmrc-service-navigation-language-select__list">
+      <li class="hmrc-service-navigation-language-select__list-item">
+        <a href="{{ params.en.href }}"
+          hreflang="en"
+          lang="en"
+          rel="alternate"
+          class="govuk-link govuk-service-navigation__link"
+          data-journey-click="link - click:lang-select:English">
+          <span class="govuk-visually-hidden">Change the language to English</span>
+          <span aria-hidden="true">ENG</span>
```
oscarduignan: @shabana-ali suggestion for this so that the visual label is a part of the accessible label is a good one I think
```suggestion
          ENG<span class="govuk-visually-hidden"> - Change the language to English</span>
```
should the dash be an endash (`-`) or a emdash (`–` like the separator we use in urls?)

app/app.js
JoPintoPaul: Thanks the explanation @oscarduignan !

src/components/service-navigation-language-select/template.njk
shabana-ali: I think emdash https://design.tax.service.gov.uk/hmrc-content-style-guide/#dashes

src/components/service-navigation-language-select/jsdom.test.js
```diff
@@ -0,0 +1,56 @@
+/* eslint-env jest */
+
+import axe from '../../../lib/axe-helper';
+
+import { render, getExamples } from '../../../lib/jest-helpers';
+
+const examples = getExamples('service-navigation-language-select');
+
+describe('New Tab Link', () => {
```
JoPintoPaul:
```suggestion
describe('Service Navigation Language Select', () => {
```

oscarduignan: good catch!

src/components/service-navigation-language-select/_service-navigation-language-select.scss
oscarduignan, line 1: I've removed the commented out CSS that was copied from the prototype, but not thoroughly reviewed the other CSS

oscarduignan: if we're happy with the CSS rules we have here for now then we should do a bit of cross browser testing

src/components/service-navigation-language-select/template.njk
```diff
@@ -0,0 +1,39 @@
...
+        <span aria-current="true">ENG</span>
+      </li>
+      <li class="hmrc-service-navigation-language-select__list-item">
+        <a href="{{ params.cy.href }}"
+          hreflang="cy"
+          lang="cy"
+          rel="alternate"
+          class="govuk-link govuk-service-navigation__link"
+          data-journey-click="link - click:lang-select:Cymraeg">
+          CYM<span class="govuk-visually-hidden"> – Newid yr iaith ir Gymraeg</span>
```
oscarduignan, line 14: @timsb I've updated these using @shabana-ali suggestion 🙌

src/govuk-prototype-kit.config.json
```diff
@@ -99,6 +99,10 @@
       "macroName": "hmrcReportTechnicalIssue",
       "importFrom": "hmrc/components/report-technical-issue/macro.njk"
     },
+    {
+      "macroName": "hmrcServiceNavigationLanguageSelect",
+      "importFrom": "hmrc/components/service-navigation-language-select/macro.njk"
+    },
```
oscarduignan, line 105: worth a sanity check maybe to install this into a prototype and ensure the new macro is usable

oscarduignan:
flakey test visual regression test not waiting after interacting with page for expected element to appear

CleanShot 2025-06-20 at 11 53 32@2x

will just try rerunning for now - but should be fixed 😓

src/govuk-prototype-kit.config.json
oscarduignan: locally built package installed fine in a new prototype, and was able to copy paste the example from description and have it display correctly

CleanShot 2025-06-20 at 11 56 03@2x

oscarduignan:
rebased on top of govuk-frontend@5.11.0 visual differences are just around the logo and baseline of text in service nav shifting down a pixel - the updated examples are just ones that I've added with the service navigation change, because existing examples were updated with the govuk-frontend@5.11.0 changes

examples of the diffs:

CleanShot 2025-06-30 at 11 50 38@2x

CleanShot 2025-06-30 at 11 52 55@2x

src/components/service-navigation-language-select/_service-navigation-language-select.scss
```diff
@@ -0,0 +1,67 @@
...
+
+.govuk-service-navigation:has(.hmrc-service-navigation-language-select) {
+  .govuk-width-container {
+    display: flex;
+  }
+}
+
+.govuk-service-navigation .hmrc-service-navigation-language-select {
+  display: inline;
+  flex: 0 0 auto;
```
oscarduignan, line 39: we think this was for ie11 but do we need it?

src/components/service-navigation-language-select/_service-navigation-language-select.scss
```diff
@@ -0,0 +1,67 @@
...
+  a,
+  [aria-current] {
+    padding: 0.3125em;
+  }
+}
+
+// TODO the styles below this line were copied verbatim from prototype, not final
+// Tim B had some ideas for refactoring it because of changes coming from GDS soon
+
+.govuk-service-navigation:has(.hmrc-service-navigation-language-select) {
```
oscarduignan: would it be better to have a class akin to .hmrc-header.hmrc-header--with-additional-navigation

oscarduignan: looking at this again while Tim is off, feels like we could say "add this extra class to the container" but also keep the `:has()` so that if someone forgets to add it then we it should still display correctly for modern browsers

```suggestion
.govuk-service-navigation.hmrc-service-navigation--with-language-select,
.govuk-service-navigation:has(.hmrc-service-navigation-language-select) {
```

where without it applying, you see the following

Image

Image

which might be usable enough, considering how few people it would apply to, and only if people forgot to add the class, which we could mostly avoid being possible in play-frontend-hmrc when people were adding the language select

oscarduignan: I'm not sure if we should do anything fancy in hmrc-header to introspect the service nav params and add the class if it's missing

oscarduignan: I've updated the reference images, but not sure I'll do anymore poking before Tim is back - not got any ideas for other approaches to laying this out

src/components/header/template.njk
```diff
@@ -93,6 +94,9 @@
     {{ govukPhaseBanner(params.phaseBanner) }}
   </div>
 {% endif %}
+{% if params.serviceNavigation %}
+    {{ govukServiceNavigation(params.serviceNavigation) }}
+{% endif %}
```
oscarduignan, line 99: just realised that I placed this in the wrong spot, service navigation should go above phase banner - it is correct in the prototype and the design system guidance, not sure how I missed it

https://design-system.service.gov.uk/components/phase-banner/

src/components/service-navigation-language-select/_service-navigation-language-select.scss
```diff
@@ -0,0 +1,74 @@
...
+    padding: 0.3125em;
+  }
+}
+
+.govuk-service-navigation:has(.hmrc-service-navigation-language-select),
+.hmrc-service-navigation--with-language-select { // if you need to support browsers without support for :has()
+  .govuk-width-container {
+    display: flex;
+  }
+}
```
oscarduignan: @timsb I was tidying this up for merging and I realised I forgot what we decided on this in terms of if it was ok to depend on browser support of :has() where I think neither of us could think of an alternative that didn't involve having to add a class to parent. Is it worth having this class `hmrc-service-navigation--with-language-select` available and part of the pattern for language select in case a service has a lot of users who are using a browser that don't have support for has()? If this isn't active, then it's still usable enough I think visually - it just flows to another line:

Image

the issue with having this extra class I can think of, is that people could add or forget to (though we would be able to protect against this in play-frontend-hmrc somewhat), which might mean that if some one was an old browser then the language select might move around if moving between services that had added the class or not - which makes me wonder if consistently a bit odd looking is better than chance of inconsistently how it was intended to look and not 

what do you think? I added the class just as an example, to make sure it was clear what I was talking to, but can easily remove it before we merge this 🙏 in use it would mean

CHANGELOG.md
oscarduignan, line 1: updated this to describe the change to hmrc banner and how to revert-  guess this release will be important one to announce

src/components/service-navigation-language-select/_service-navigation-language-select.scss
timsb: I think I had decided on using just the class rather than the `:has` pseudo class. I think there _are_ some downsides for people using prototypes with this approach - they have to manually add the class and may forget to across all pages and/or interlinked prototype journeys.

As you say you can protect against this in play-frontend-hmrc where it really matters. How reliably could you add the class? Would there be some work from teams to get the class added automatically or would you just detect if they're using the service nav and language toggle and add it automatically?

The positives for using just a class is that in prototypes designers will have the flexibility to add or remove it as they like if they want to experiment with different designs whereas if we use `:has` as well, it'll be harder for them to override.

oscarduignan: we can really reliably add it - and I guess it will be really visible if they don't add it so shouldn't make it to prod

oscarduignan: I'll remove the :has 🙌

oscarduignan: does the class name look good to you that I've used? @timsb

timsb: I think it's odd that we're prefixing the 'thing' with 'hmrc' when really it's just the govuk Service nav but it's what gov.uk DS suggests we do (https://design-system.service.gov.uk/get-started/extending-and-modifying-components/#small-modifications-to-components). I wondered whether making the modifier part of the class name more generic such as `--push-to-end` or something like that but I think we only want people using this for the language switcher. I think it's a good, descriptive class name so yeah I'm happy with it.

src/components/header/header.yaml
```diff
@@ -641,3 +649,66 @@ examples:
...
+                    CYM<span class="govuk-visually-hidden"> – Newid yr iaith ir Gymraeg</span>
+                  </a>
+                </li>
+              </ul>
+            </nav>
+    languageToggle:
+      en:
+        href: "/components/header/with-language-toggle-english/preview"
+      cy:
+        href: "/components/header/with-language-toggle-welsh/preview"
```
oscarduignan, line 685: @timsb I've updated the example here so it includes the class that's needed

I wish that we could do object destructuring in nunjucks to make it easier to wrap another component

because it feels like it would be so much nicer to have a hmrcServiceNavigation fork that had language select params in it and handled based on that the rendering of language select in the right place and adding of the class

but I know it will be like hmrcFooter where we have to keep track of all the params and pass them through https://github.com/hmrc/hmrc-frontend/blob/main/src/components/footer/template.njk#L18-L36 which makes it easy to miss new params being added during updates and just makes breakages easier during updates to govuk-frontend :(

oscarduignan: https://github.com/mozilla/nunjucks/issues/1455#issuecomment-1904221956 seeing this from Ed makes me wonder if it's worth having a look at nhs-frontend to see if they are doing anything copyable

oscarduignan: nothing in nhsuk-frontend, so guess it's more something from prototypes, if people find it challenging to use, then maybe it's just worth owning the extra maintenance for teams of a fork of service navigation with extra language select param and a bit of verbosity mapping params in it's template

a side effect of that would be that it would be quite easy for us to tweak the service navigation for all teams in the future if we needed to 🤷‍♂️

backstop_data/bitmaps_reference/backstop_default_service-navigation-language-select_default_example_using_rebrand_0_document_0_phone.png
timsb: Pretty sure I fixed this as part of f9f2cd689199ee65c5b751599ce4d877bbced189

src/components/header/header.yaml
timsb: Destructuring would be a nicer way of handling this. Thanks for looking into nhsuk-frontend. We'll keep an eye on this release then and see how people are doing.
It might be an issue that creeps up over time as new people join and weren't here for the initial release

src/components/service-navigation-language-select/_service-navigation-language-select.scss
```diff
@@ -0,0 +1,73 @@
...
+  display: inline-block;
+
+  &:first-child::after {
+    content: "";
+    display: inline-block;
+    position: relative;
+    top: 0.1875em;
+    height: 1em;
+    // stylelint-disable-next-line number-max-precision -- gives a 1.5px border width
+    border-right: 0.09375em solid  $govuk-text-colour;
```
timsb: Really minor but couldn't ignore it once I'd spotted it
```suggestion
    border-right: 0.09375em solid $govuk-text-colour;
```

src/components/service-navigation-language-select/_service-navigation-language-select.scss
timsb: I looked into this and found the default is `0 1 auto` so we do need this to stop shrinking. We could just use `flex-shrink` but I think I prefer setting all the properties out even if they are the defaults

PatrykRudnicki (approved):

PR #467, opened 2025-06-23
PLATUI-3788 try-catch on initialisation

TimothyFothergill:
# Try-Catch on Component Initialisation

**Bug fix**

When we initialise a component, if one fails, but there are multiple components that should be initialised on the same page, then they all silently don't get initialised. This ensures that if one component fails, we log it to console, but we allow the rest of the components to initialise.

## Checklist

### DIFF
--- a/src/all.js
+++ b/src/all.js
@@ -6,13 +6,17 @@ import SessionActivityService from './components/timeout-dialog/session-activity
 import HmrcPrintLink from './components/hmrc-print-link/hmrc-print-link';
 
 function initAll() {
+  function initError(componentName, message) {
+    // eslint-disable-next-line no-console
+    return console.error(`hmrc-frontend - ${componentName} failed to initialise: ${message}`);
+  }
+
   const $AccountMenuSelector = '[data-module="hmrc-account-menu"]';
   if (document.querySelector($AccountMenuSelector)) {
     try {
       new AccountMenu($AccountMenuSelector).init();
     } catch ({ message }) {
-      // eslint-disable-next-line no-console
-      console.error(`hmrc-frontend - AccountMenu failed to initialise: ${message}`);
+      initError('AccountMenu', message);
     }
   }
 
@@ -21,8 +25,7 @@ function initAll() {
     try {
       new HmrcPrintLink($HmrcPrintLink, window).init();
     } catch ({ message }) {
-      // eslint-disable-next-line no-console
-      console.error(`hmrc-frontend - HmrcPrintLinks failed to initialise: ${message}`);
+      initError('HmrcPrintLink', message);
     }
   });
 
@@ -34,8 +37,7 @@ function initAll() {
     try {
       new TimeoutDialog($TimeoutDialog, sessionActivityService).init();
     } catch ({ message }) {
-      // eslint-disable-next-line no-console
-      console.error(`hmrc-frontend - TimeoutDialog failed to initialise: ${message}`);
+      initError('TimeoutDialog', message);
     }
   }
 
@@ -44,8 +46,7 @@ function initAll() {
     try {
       new UserResearchBanner($UserResearchBanner).init();
     } catch ({ message }) {
-      // eslint-disable-next-line no-console
-      console.error(`hmrc-frontend - UserResearchBanner failed to initialise: ${message}`);
+      initError('UserResearchBanner', message);
     }
   }
 
@@ -55,7 +56,7 @@ function initAll() {
       new BackLinkHelper($BackLink, window, document).init();
     } catch ({ message }) {
       // eslint-disable-next-line no-console
-      console.error(`hmrc-frontend - BackLinkHelper failed to initialise: ${message}`);
+      initError('BackLinkHelper', message);
     }
   });
 }
--- a/src/all.js
+++ b/src/all.js
@@ -6,27 +6,28 @@ import SessionActivityService from './components/timeout-dialog/session-activity
 import HmrcPrintLink from './components/hmrc-print-link/hmrc-print-link';
 
 function initAll() {
-  function initError(componentName, message) {
-    // eslint-disable-next-line no-console
-    return console.error(`hmrc-frontend - ${componentName} failed to initialise: ${message}`);
+  function logAndIgnoreErrors(init) {
+    try {
+      init();
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      return console.error('hmrc-frontend component initialisation failed', error);
+    }
+    return true;
   }
 
   const $AccountMenuSelector = '[data-module="hmrc-account-menu"]';
   if (document.querySelector($AccountMenuSelector)) {
-    try {
+    logAndIgnoreErrors(() => {
       new AccountMenu($AccountMenuSelector).init();
-    } catch ({ message }) {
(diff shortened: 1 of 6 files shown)


src/all.js
```diff
@@ -6,32 +6,58 @@ import SessionActivityService from './components/timeout-dialog/session-activity
 import HmrcPrintLink from './components/hmrc-print-link/hmrc-print-link';
 
 function initAll() {
+  function initError(componentName, message) {
+    // eslint-disable-next-line no-console
+    return console.error(`hmrc-frontend - ${componentName} failed to initialise: ${message}`);
+  }
```
oscarduignan: I think initError could be easily confusable for "initialise an error" rather than "handle an error during initialisation"
what about renaming this `logError` and I reckon rather than logging the message we could log the whole error, so: `catch(error)` and 

```
console.error(`hmrc-frontend - ${componentName} failed to initialize`, error)
```

oscarduignan: since we're just using a string and the error, you could wrap it more 

```
logAndIgnoreErrors(() => {
  new AccountMenu($AccountMenuSelector).init();
});
```

I guess that the error will include something that indicates that it occurred in AccountMenu

```
function logAndIgnoreErrors(init) {
   try {
      init();
    } catch (error) {
      return console.error(`hmrc-frontend component initialisation failed`, error);
    }
}
```

then I think inside components, we could throw better exceptions if they aren't clear enough - like
```
throw new Error("Missing X argument required to initialise hmrc timeout dialog", element);
throw new Error("Can't init second timeout warning on a single page", element);
// etc
```

src/all.jsdom.test.js
```diff
@@ -32,5 +32,22 @@ describe('Init All', () => {
...
+            data-timeout="60"
+            data-countdown="55"
+            data-keep-alive-url="?keepalive"
+            data-sign-out-url=""
+            data-synchronise-tabs="true"/>
+        `; // broken configuration of timeout-dialog will error
+      document.body.appendChild(brokenTimeoutDialog.firstChild);
+      HMRCFrontend.initAll();
+      expect(BackLinkHelper.prototype.init).toHaveBeenCalled();
+    });
```
oscarduignan: can we `document.createElement('meta')` and `setAttribute('content', 'hmrc-timeout-dialog')` on it like that for consistency with other examples) and then append to the document.head?

TimothyFothergill: For clarity, I can do this, but also do another `setAttribute('data-sign-out-url', '')`. This way we know that the component will be set up with nothing in that field, not a default, so it'll definitely fail.

TimothyFothergill: Opted to not add the second setAttribute, as there is no default value there 🙌🏻

src/all.js
```diff
@@ -6,32 +6,52 @@ import SessionActivityService from './components/timeout-dialog/session-activity
...
 
 function initAll() {
+  function logAndIgnoreErrors(init) {
+    try {
+      init();
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      return console.error('hmrc-frontend component initialisation failed', error);
+    }
+    return true;
```
oscarduignan:
```suggestion
      console.error('hmrc-frontend component initialisation failed', error);
    }
```

src/all.jsdom.test.js
```diff
@@ -32,5 +32,14 @@ describe('Init All', () => {
       HMRCFrontend.initAll();
       expect(BackLinkHelper.prototype.init).toHaveBeenCalledTimes(2);
     });
+
+    it('should initialise correctly configured elements found on the page', () => {
+      addBackLinkToPage();
+      const brokenTimeoutDialog = document.createElement('meta');
+      brokenTimeoutDialog.setAttribute('content', 'hmrc-timeout-dialog'); // broken configuration of timeout-dialog will error
```
oscarduignan: sorry, I think we will need the other attributes - or at least `meta[name="hmrc-timeout-dialog"` otherwise it won't try to init timeout dialog and test would pass because no error during initialisation

oscarduignan: could maybe spy on the console.error during test to check that got error logged to show it attempted to init timeout dialog

src/all.js
oscarduignan: small one, just because we're not doing anything with what we return so probably no point

oscarduignan: govuk return stuff because they return components that were initialised, because I think they expose APIs to interact with them elements afterwards - but we just init and discard - and any other stuff people would need to init themselves for

oscarduignan:
CI error is the flakey VRT test again that isn't waiting for menu to be open after clicking it - this happened on another PR for me, maybe the number of VRT tests we have now with rebranded state is causing it to be more likely

CleanShot 2025-06-23 at 15 54 26@2x

could check it out as part of this, or we can spin up a bug ticket - I was able to rerun tests on my PR but with it happening twice in a handful of builds it seems maybe likely enough that it could hit a release build. I think we had a way to make the vrt tests wait, maybe just needs bit more time or something - or we missed a scenario

src/all.js
TimothyFothergill: Completely fair, removed returns because yeah, there's no point 🙌🏻

oscarduignan (approved):

src/components/account-header/account-header.yaml
```diff
@@ -142,6 +142,7 @@ examples:
       classes: 'the-header-classes'
   visualRegressionTests:
     backstopScenarioOptions:
+      delay: 2000
```
oscarduignan, line 145: I'm not sure if the delay happens before or after the "beforeTakingScreenshot" interactions - if it's after then we're all good, if it's before then we might need to add a waitFor just after clicking actions are taken in the alternateStates
```
      account-menu-dropdown-open:
        beforeTakingScreenshot:
          - click: ".hmrc-account-menu__link[aria-expanded='false']"
          - waitFor: ".main-nav-is-open"
 ```

oscarduignan: actually, it's that one that it's failing on - so maybe the waitFor isn't enough or is using the wrong selector

oscarduignan: it's odd, because those beforeTakingScreenshot actions happen onready so js has fully loaded
 
 https://github.com/hmrc/hmrc-frontend/blob/c48d922d7d808450aa5006e489d85ec72f420d9e/backstop_data/engine_scripts/puppet/onReady.js#L4
 
 and waitFor is just [delegating to standard waitForSelector ](https://github.com/hmrc/hmrc-frontend/blob/c48d922d7d808450aa5006e489d85ec72f420d9e/backstop_data/engine_scripts/puppet/beforeTakingScreenshotHelper.js#L19-L20) - maybe all that points to .main-nav-is-open happening too early such that sometimes it is successful before nav is visibly open

oscarduignan: dunno, seems like the right selector 🤷 from looking at https://github.com/hmrc/hmrc-frontend/blob/c48d922d7d808450aa5006e489d85ec72f420d9e/src/components/account-menu/account-menu.js#L62-L70 think if it happens again after the delay, probably can avoid it by adding a short delay after waiting for main-nav-is-open

PR #469, opened 2025-07-01
Nojira update stylelint

PatrykRudnicki:
# Descriptive Title

Fixes # (issue)

## Checklist

* [X] I've read the [CONTRIBUTING](../CONTRIBUTING.md) guidance, including next steps and [expected response](../CONTRIBUTING.md#when-can-i-expect-someone-to-look-at-my-external-contribution) from the owners
* [X] I've added appropriate unit tests, and run all [unit tests](../CONTRIBUTING.md#unit-tests)
* [X] I've run the [visual regression test using Backstop](../CONTRIBUTING.md#visual-regression-tests) and updated the images if needed
* [X] I've run `npm version minor` to update the version in [package.json](../package.json) and [package-lock.json](../package-lock.json)
* [X] I've updated the [CHANGELOG](../CHANGELOG.md)

### DIFF
--- a/src/components/header/_header.scss
+++ b/src/components/header/_header.scss
@@ -1,4 +1,4 @@
-// stylelint-disable
+// stylelint-disable max-nesting-depth, number-max-precision -- deferred during move to stylelint-config-gds
 $govuk-header-link-underline-thickness: 3px;
 
 .hmrc-header.hmrc-header--with-additional-navigation {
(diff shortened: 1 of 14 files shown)


src/components/header/_header.scss
```diff
@@ -1,3 +1,4 @@
+// stylelint-disable
```
oscarduignan: @PatrykRudnicki would it be possible to list the specific rules we're disabling (unless there's lots) for these, I think disabling them for the whole file is fine but it could use a bit more context, so maybe something like:
```
/* stylelint-disable x, y, z -- deferred during move to stylelint-config-gds */
```
that + the comment will make it a bit more clear what and why stuff is disabled I think

oscarduignan (approved):

PR #471, opened 2025-07-22
PLATUI-3856 script to check compatibility of frontend versions

TimothyFothergill:
# Add script to check compatibility of frontend versions

**New feature**

Adds a script that checks the required version of govuk-frontend with hmrc-frontend.

- This should only impact prototypes, as hmrc services pull hmrc-frontend in via play-frontend-hmrc, which bundles govuk-frontend with it

## Checklist


lib/check-govuk-and-hmrc-frontend-versions.js
```diff
@@ -0,0 +1,39 @@
...
+  -------------------------------------------
+          \\   ^__^ 
+           \\  (oo)_______
+              (__)       )\\/\\
+                  ||----w |
+                  ||     ||
+  - PlatMooI the PlatUI Helper Cow
+  `);
+
+  process.exit(1);
```
oscarduignan: will this interrupt the installation or will it continue?

package.json
```diff
@@ -21,7 +21,8 @@
     "test:backstop-approve": "gulp backstop:approve",
     "docs:generate-decision-log-listing": "adr-log -d ./docs/adr -i ./docs/adr/index.md -e template.md",
     "audit": "better-npm-audit audit",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "postinstall": "node lib/check-govuk-and-hmrc-frontend-versions.js"
```
oscarduignan: we might need to copy this into the package, so it's available when installing - or is lib being included?

the last version we had it at top level with a gulp target to copy it into package

https://github.com/hmrc/hmrc-frontend/commit/2b846b5bde540fa463d69098a64d9402fb4e2acb#diff-25789e3ba4c2adf4a68996260eb693a441b4a834c38b76167a120f0b51b969f7L29

what do you think about moving it to the top level and making it "check-compatibility-with-govuk-frontend-version.js"

lib/check-govuk-and-hmrc-frontend-versions.js
TimothyFothergill: This is done as a postinstall, so installation should have finished. However, it would likely fail builds. I suppose when it gets to the point that they're trying to build, is it better to error (because there's a requirement unsatisfied) or just say "it won't work as you expect"?

lib/check-govuk-and-hmrc-frontend-versions.js
oscarduignan, line 1: I think we should have a test for this, here's how we tested it before with subprocesses - there might be better ways in current version of node we're on - maybe around creating temporary folders, worth checking what latest recommendation from jest is for this kind of thing

https://github.com/hmrc/hmrc-frontend/commit/2b846b5bde540fa463d69098a64d9402fb4e2acb#diff-053cc2cf2011477165296db1508ce5a2db5c44bc0284b5d0a82891034f8e2840

oscarduignan:
in the description we have "This should only impact prototypes, as hmrc services pull hmrc-frontend in via play-frontend-hmrc, which bundles govuk-frontend with it" but it will impact anyone who pulls in hmrc-frontend via npm I guess - wonder if we should have ability to skip the compatibility check in case anyone wants to disable it

package.json
TimothyFothergill: Sounds good to me! Will get that moved to the root of the project and renamed

lib/check-govuk-and-hmrc-frontend-versions.js
```diff
@@ -0,0 +1,39 @@
+const fs = require('fs');
+const path = require('path');
+
+const projectDir = process.env.INIT_CWD;
+const hmrcFrontendDir = path.join(projectDir, 'node_modules/hmrc-frontend/');
+let govukFrontendVersion = '';
+
+try {
+  govukFrontendVersion = JSON.parse(fs.readFileSync(path.join(hmrcFrontendDir, 'package.json'), 'utf8')).dependencies['govuk-frontend'].replace(/[^0-9.]/g, '');
```
oscarduignan: I know we were doing it in the old script, but I'm not sure why, I think we could use require() to import this

readFileSync won't cache the result, but since this script is just getting run in a subprocess or tests that doesn't seem like it's much of an issue

that said, probably not worth refactoring

TimothyFothergill:
> in the description we have "This should only impact prototypes, as hmrc services pull hmrc-frontend in via play-frontend-hmrc, which bundles govuk-frontend with it" but it will impact anyone who pulls in hmrc-frontend via npm I guess - wonder if we should have ability to skip the compatibility check in case anyone wants to disable it

Debated this too, but in terms of who usually consumes hmrc-frontend directly it is prototype developers, but I'm all for a skip, personally!

lib/check-govuk-and-hmrc-frontend-versions.js
oscarduignan: ah on this, we're parsing the hmrc-frontend govuk-frontend version I realise, I wonder though if we want to track the version with which we know that installing govuk-frontend alongside hmrc-frontend in a prototype will break (due to breaking changes we made or govuk-frontend made, like dist/ paths and where we unintentionally did a breaking change by implementing deprecation notice too early)

TimothyFothergill: The current approach above allows us to get away with only needing to make the one change in package.json (well and package-lock.json). Would tracking a known breaking version require more manual work to keep the check updated? I noticed in the previous script, we had a manually-updated list: https://github.com/hmrc/hmrc-frontend/blob/6c3847de33a25c02d1a737c2993b1b568c1afcd8/check-compatibility.js#L16-L26.

lib/check-govuk-and-hmrc-frontend-versions.js
```diff
@@ -0,0 +1,39 @@
...
+} catch (err) {
+  process.exit(0); // Exit if no package.json is found under project_path/node_modules/hmrc-frontend
+}
+
+const govukPrototypeKitPackagePath = path.join(projectDir, 'package.json');
+const govukPrototypeKitConfig = JSON.parse(fs.readFileSync(govukPrototypeKitPackagePath, 'utf8'));
+const currentGovukFrontendVersion = govukPrototypeKitConfig.dependencies['govuk-frontend'].replace(/[^0-9.]/g, '');
```
oscarduignan: I think some of the variable names confused me a bit

in my head there's

- the prototype (or application or project)
- the govuk-prototype-kit library which prototypes depend on 
- the hmrc-frontend library

where from skimming the variables I would assume that currentGovukFrontendVersion was version that govuk-prototype-kit library depends on but I think it's actually the project

I think as well we should probably call it project unless we are checkign if we're in prototype context and only preceding if we are, in that case we can say we've refined scope to prototype and refer to it like that - so that we don't accidentally imply that we're only doing this in a more limited scope than we are through our names

I am totally over engineering it, but my reasoning for it, is because this script will probably sit untouched for a while so if we can make it as clear as possible then that will help future maintainers a lot to understand what it's doing and why it's useful (or maybe not anymore, for example, if every prototype in hmrc has upgraded to a version of govuk-frontend past the last one we are fatally incompatible with then 🤷‍♂️ maybe can get rid of it)

I would be interested to hear what @JoPintoPaul and @PatrykRudnicki think about my suggestions, I might be going to far

lib/check-govuk-and-hmrc-frontend-versions.js
oscarduignan: I don't know how postinstall scripts chain, if it has the chance if one fails to interrupt others from other dependencies

oscarduignan: in my head, completing check successfully regardless of outcome might be best way to have the exitcode and then it's just giving a visible warning but not interrupting anything because it could be fine to be incompatible depending on how they are using it

lib/check-govuk-and-hmrc-frontend-versions.js
TimothyFothergill: Happy to update variable names - Happy to wait for extra feedback from the team. It might be more appropriate as:
`projectPackagePath` (projectDir/package.json)
`projectPackageJson` (JSON.parse(projectPackagePath))
`currentGovukFrontendVersionInProject` (projectPackageJson.dependencies['govuk-frontend'])

package.json
oscarduignan: or maybe, because I love verbose stuff "warn-about-potential-incompatibility-with-govuk-frontend-version.js" xD dunno

package-lock.json
oscarduignan, line 1: I can't see the diff in github, but it seems like a lot of changes when we haven't added any dependencies. +7k and -5k

TimothyFothergill: That is odd! I'll double check what happened there.

lib/check-govuk-and-hmrc-frontend-versions.js
TimothyFothergill: For clarity for any future maintainers who might read this, spoke to Oscar directly about the reasoning behind using `process.exit(1)`. Without erroring, npm by default silences any output for non-failing builds (log, warn, error etc). By erroring out, we get the logs out with the npm logs.

lib/check-govuk-and-hmrc-frontend-versions.js
oscarduignan: I think we could just have 1 version, like "to our knowledge, this version of hmrc-frontend won't work with versions of govuk-frontend before this version" and update it when we make a breaking change like use something only available in new version of govuk-frontend or something like that

oscarduignan: where I guess the main way we've seen stuff break is things that impact styles or path changes

tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
```diff
@@ -0,0 +1,99 @@
...
+      child.on('exit', (code) => {
+        try {
+          expect(code).toBe(0);
+          done();
+        } catch (err) {
+          done(err);
+        }
+      });
+    });
+  });
```
oscarduignan: does this need to have the hmrc-frontend and govuk-frontend dependencies?

tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
```diff
@@ -0,0 +1,99 @@
...
+
+  describe('Installing inside of a Prototype kit project', () => {
```
oscarduignan, line 54: do these need to have govuk-prototype-kit installed?

oscarduignan: (to be in the context of a prototype kit project)

tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
```diff
@@ -0,0 +1,99 @@
...
+        const child = exec(`node ${scriptPath}`);
+        child.on('exit', (code) => {
+          try {
+            expect(code).toBe(0);
+            done();
+          } catch (error) {
+            done(error);
+          }
+        });
+      });
```
oscarduignan: for this, do we need to copy the failing createMockPackage from above, and then assert that it's 0 where if we didn't have HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK it would be 1 (based on the evidence of expected failing previous test)

tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
```diff
@@ -0,0 +1,99 @@
...
+
+    describe('Installing with an incompatible version', () => {
+      it('should exit the process with code 1', (done) => {
+        createMockPackage({
+          dependencies: {
+            'govuk-frontend': '4.4.0',
+            'hmrc-frontend': '6.79.0',
+          },
+        });
+        const child = exec(`node ${scriptPath}`, { env: process.env.INIT_CWD });
```
oscarduignan: just noticed you pass this explicitly from above, I wouldn't set it as an environment variable then in this file, but a constant

tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
oscarduignan, line 1: nodejs sub process stuff is so wordy, wonder if we could wrap it to reduce the boilerplate and make tests easier to parse, something that lets us do:

```js
const exitCode = await runCompatibilityCheckAgainstTheMockedPackage()
```

maybe something like:
```js
function runCompatibilityCheckAgainstTheMockedPackage() {
  const proc = spawn("node check-compatibility.....js")
  return new Promise((resolve) => {
    proc.on('exit', (code) => {
        resolve(code);
    });
  })
}
```

oscarduignan: could go further I guess 

- given, `createMockedPackage` => `givenPackageJsonOf({ ... }`
- when, `const exitCode = await runCompatibilityCheck()`
- then , `expect(exitCode).toBe(0)`

check-compatibility-with-govuk-frontend-version.js
```diff
@@ -0,0 +1,37 @@
+if (process.env.HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK === 'true') {
+  process.exit(0);
+}
+const fs = require('fs');
+const path = require('path');
+
+// To our knowledge hmrc-frontend won't work with versions below this.
+// This is because of a change to directory structure (/dist/).
+const minimumGovukFrontendVersion = '5.1.0';
```
oscarduignan: I think it might be higher than this now - I think it's 5.4.0 because we implemented a deprecation notice too early

https://hmrcdigital.slack.com/archives/C19R6A9L7/p1737974982425489

check-compatibility-with-govuk-frontend-version.js
```diff
@@ -0,0 +1,37 @@
...
+
+const projectDir = process.env.INIT_CWD;
+const projectPackagePath = path.join(projectDir, 'package.json');
+
+const projectPackageJson = JSON.parse(fs.readFileSync(projectPackagePath, 'utf8'));
+const currentVersionInProject = projectPackageJson.dependencies['govuk-frontend'].replace(/[^0-9.]/g, '');
+if (currentVersionInProject >= minimumGovukFrontendVersion) {
```
oscarduignan: I don't think we're comparing the semver strings correctly here, string comparison will look correct some of the time, but it's just comparing digit by digit - not part by part, so for example:

```js
("9.2.0" > "9.1.0") == true
("10.0.0" > "9.0.0") == false
("10.99.0" > "10.19.0") == true
("10.100.0" > "10.19.0") == false
```

would be good to have some test cases like these, and then I think we'd need to split the semver string and compare parts one at a time

oscarduignan: here's an example (in scala) of how to compare that I had to do recently https://github.com/hmrc/platui/pull/10/files#diff-e803cb6aad6d0e19a39f9224e7491f0f249dcc9ee92844cfd0cf37d2b18b5dfaR61-R69

some logic like this to see if X semver is greater than Y semver
split the parts of the semver strings, for each part, if X.major == Y.major then compare minor, otherwise return X.major > Y.major

oscarduignan: we also could just use https://www.npmjs.com/package/semver - since it's from npm - don't feel too bad about adding a new dependency on it for this 🤷 maybe - though when I say it out loud I'm like "noo"

oscarduignan: apparently because of the fact that 0 == false in js, you can write it very succinctly! So don't need to pull in semver dep I think, if we are happy how we're handling edge case versions which think we probably given we're comparing to one specific version

```js
function semverGreaterThan(a, b) {
  const [aMaj, aMin, aPat] = a.split('.').map(Number);
  const [bMaj, bMin, bPat] = b.split('.').map(Number);
  return ((aMaj - bMaj) || (aMin - bMin) || (aPat - bPat)) > 0;
};
```

tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
```diff
@@ -0,0 +1,132 @@
...
+
+  beforeAll(() => {
+    process.env.HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK = '';
+  });
+
+  describe('Installing outside of a Prototype kit project', () => {
+    it('should exit the process with code 0', (done) => {
+      createMockPackage({
+        dependencies: {
+          'govuk-frontend': '5.11.0',
```
oscarduignan: I think here we should use an incompatible version for the test to show that "even though using incompatible version, because you're outside prototype kit, it won't do anything"

```suggestion
          'govuk-frontend': '4.4.0',
```

tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
```diff
@@ -0,0 +1,132 @@
...
+  afterAll(() => {
+    unlink(mockPackagePath, () => {});
+  });
+
+  beforeAll(() => {
+    process.env.HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK = '';
+  });
+
+  describe('Installing outside of a Prototype kit project', () => {
+    it('should exit the process with code 0', (done) => {
```
oscarduignan:
```suggestion
    it('should not run the check', (done) => {
```

check-compatibility-with-govuk-frontend-version.js
```diff
@@ -0,0 +1,49 @@
+if (process.env.HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK === 'true') {
+  process.exit(0);
+}
+const fs = require('fs');
+const path = require('path');
+
+// To our knowledge hmrc-frontend won't work with versions below this.
+const minimumGovukFrontendVersion = '5.4.0';
```
oscarduignan: @TimothyFothergill sorry was just circling back to this in my hungover state, I guess we might need to make the semver check be greater than or equal to so that someone on 5.4.0 is ok

check-compatibility-with-govuk-frontend-version.js
```diff
@@ -0,0 +1,49 @@
...
+
+const projectPackageJson = JSON.parse(fs.readFileSync(projectPackagePath, 'utf8'));
+
+try {
+  projectPackageJson.dependencies['govuk-prototype-kit'].replace(/[^0-9.]/g, '');
+} catch (err) {
+  process.exit(0); // Exit if govuk-prototype-kit is not found in package.json
+}
+
+const currentVersionInProject = projectPackageJson.dependencies['govuk-frontend'].replace(/[^0-9.]/g, '');
```
oscarduignan: @TimothyFothergill I think we need to to exit(0) if govuk-frontend isn't found in package.json because I guess that it's totally legit to have no direct dependency on govuk-frontend, probably worth adding a test for that too

oscarduignan: apologies for coming back with more stuff, I guess this is one of those ones where I'm being extra persnickety because I think if we did the wrong thing we could get a flood of "what have you done, you've stopped our project building!"

in the future I feel like there should be something we can do with peerDependencies but I had a play around locally and couldn't get them to to actually warn how they say they will when incompatible version is present

check-compatibility-with-govuk-frontend-version.js
```diff
@@ -0,0 +1,49 @@
...
+const minimumGovukFrontendVersion = '5.4.0';
+
+function semverGreaterThan(a, b) {
+  const [aMaj, aMin, aPat] = a.split('.').map(Number);
+  const [bMaj, bMin, bPat] = b.split('.').map(Number);
+  return ((aMaj - bMaj) || (aMin - bMin) || (aPat - bPat)) > 0;
+}
+
+const projectDir = process.env.INIT_CWD;
+const projectPackagePath = path.join(projectDir, 'package.json');
```
oscarduignan: I think for these we have the risk of an exception being thrown too, because INIT_CWD can be undefined

wonder if we should wrap the whole file in a try catch and exit(0) on any exception so it's safe by default

check-compatibility-with-govuk-frontend-version.js
nataliecarey: I’ve been watching this PR and trying to work out whether to comment or not.

You might want to cater for release candidates e.g. 5.1.3-rc.1

I’ve got a couple of other observations that I’ll email in case they’re helpful.

oscarduignan: thank you Natalie! you are a legend 🙏

oscarduignan:
having a think about overall safe structure for this that I'm imagining as a target is something like this sparse outline

```js
// prototypes load and compile a subset of hmrc-frontend styles using their
// own version of govuk-frontend, if their version of govuk-frontend is 
// lower than 5.4.0 then the compilation of styles will fail with an error
// that can be hard for interaction designers to debug, we would like to 
// give them a warning on installation so that they don't get stuck later 
// when trying to use the hmrc-frontend styles.

try {
    // skip the compatibility check on install using the following env var:
    // 
    //    HMRC_FRONTEND_SKIP_GOVUK_FRONTEND_COMPATIBILITY_CHECK=1
    //
    if (shouldSkipGovukFrontendCompatibilityCheck()) { process.exit(0) }
    
    if (withinPrototypeAndUsingIncompatibleGovukFrontendVersion()) {
        console.warn("...")
        process.exit(1) // interrupts install but needed so warning reported
    }
} catch (e) {
    // we want to display a warning when hmrc-frontend is installed into a 
    // prototype which has an incompatible version of govuk-frontend, 
    // without risking interrupting install of any other projects.
}

function shouldSkipGovukFrontendCompatibilityCheck() { ... }

function withinPrototypeAndUsingIncompatibleGovukFrontendVersion() { ... }
```

oscarduignan:
and I guess questions in my head for future
- can this be done more idiomatically by moving to some usage of peerDependencies
- can we refactor how we provide styles to prototypes so that there's no chance of incompatibility

check-compatibility-with-govuk-frontend-version.js
```diff
@@ -0,0 +1,65 @@
...
+  if (
+    !projectPackageJson.dependencies
+    || !projectPackageJson.dependencies['govuk-frontend']
+    || !projectPackageJson.dependencies['govuk-prototype-kit']
+  ) {
+    process.exit(0);
+  }
+
+  const currentVersionInProject = projectPackageJson.dependencies['govuk-frontend'].replace(/[^0-9.]/g, '');
+
```
oscarduignan, line 35: it looks good to me from a safety perspective, in that the projects this should impact are prototypes so it feels like a safe thing to introduce even though it will be a breaking change for them if they are using incompatible versions

from a taste perspective, I would move all this logic into withinPrototypeKitAndUsingIncompatibleGovukFrontendVersion and not pass any arguments to it - since it's just a script and we won't be testing parts of it but as a whole I don't think there's much value in having checks split out - where withinPrototypeKitAndUsingIncompatibleGovukFrontendVersion is a nice label to describe what that code is checking, but at the moment we're checking some of that stuff thats described in the function name outside of it, which I think is an easy footgun for future

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [6.79.0] - 2025-08-06
+
+### Changed
+
+- Added a script to check the compatibility of hmrc and govuk frontends
```
oscarduignan, line 14: I think we should add an action required entry here, "if you're using a prototype that's using govuk-frontend less than 5.4.0 then installing this version of hmrc-frontend will report an error, you can update your govuk-frontend version or suppress it by setting HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK=true"

tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
oscarduignan, line 1: @TimothyFothergill just seen, I think we should have a test where INIT_PWD isn't set so we can show that it works how we expect in the absence of that (I presume it will always be set when running `npm install` but there are other package managers that might handle stuff differently)

tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
```diff
@@ -0,0 +1,185 @@
...
+  describe('Installing outside of a Prototype kit project', () => {
+    it('should not run the check', (done) => {
+      createMockPackage({
+        dependencies: {
+          'govuk-frontend': '4.4.0',
+          'hmrc-frontend': '6.79.0',
+        },
+      });
+
+      const child = exec(`node ${scriptPath}`, { env: { ...process.env, INIT_CWD: initCwd } });
```
oscarduignan: I think our use of INIT_CWD is a bit mixed up here

I would expect in my head for tests we might create a temp dir, create package.json in there, and in subprocesses where we directly run node <script> against it we pass through INIT_CWD: tempDir - where we're kind of simulating the meaningful-to-its-execution state it will get if run via npm install

but that we wouldn't be using or setting the INIT_CWD of node process running the tests

oscarduignan: I think the createMockPackage stuff is carried over from old script but I think it might be worth creating a temp dir instead that we clean up after tests run and place it in there

check-compatibility-with-govuk-frontend-version.js
```diff
@@ -0,0 +1,69 @@
...
+            \\  (oo)_______
+                (__)       )\\/\\
+                    ||----w |
+                    ||     ||
+    - PlatMooI the PlatUI Helper Cow
+    `);
+    return true;
+  }
+  return false;
+}
```
oscarduignan: sorry! I think keep the output separate from this, so can just return
```
return !semverGreaterThanOrEqualTo(currentVersionInProject, minimumGovukFrontendVersion)
```

basically so you can read the main block of code and understand what is happening, but not need to know how it is doing it, whereas without the output there you can't know what it will display that, just that it will error

oscarduignan: basically just collapsing/abstracting the conditional logic

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+- Action Required: 
+  - If you're using a prototype that's using govuk-frontend less than 5.4.0 then installing this version of hmrc-frontend will report an error. 
+  - You can update your govuk-frontend version or suppress it by setting HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK=true
+
```
```suggestion
### Action required in prototypes using govuk-frontend less than 5.4.0 
  - We now display an error on install to warn prototype maintainers of incompatibility with versions of govuk-frontend less than 5.4.0 in prototypes
  - You can update your govuk-frontend version or suppress it by setting HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK=true

```

oscarduignan: other than this one bit it all looks good to me! 🙌

package.json
```diff
@@ -1,9 +1,10 @@
 {
   "name": "hmrc-frontend",
-  "version": "6.78.0",
+  "version": "6.79.0",
   "description": "Design patterns for HMRC frontends",
   "scripts": {
     "start": "gulp dev",
+    "test:compatibility": " jest tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js",
```
oscarduignan, line 7: should we link this into another test task, so that we don't have to remember to run it if we are doing `npm test`

oscarduignan (approved):

PR #474, opened 2025-08-06
PLATUI-3856: example of not modifying test process env

oscarduignan:
just a draft to show what I mean for tim on a pr

guess it went a bit further, but it's to show the idea of creating a tmp dir during tests / not modifying process.env to set HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK for current process so it's inherited through `...procces.env` when exec'ing subprocess (took me some tapping of keyboard to get to it but basically this is the main bit, so could just pluck that - basically just "ok to inherit process.env but not to mutate it")


tasks/gulp/__tests__/check-compatibility-with-govuk-frontend-version.test.js
```diff
@@ -1,21 +1,35 @@
...
+  let tmpDir;
 
-  afterAll(() => {
-    unlink(mockPackagePath, () => {});
+  beforeEach(async () => {
+    tmpDir = await mkdtemp(path.join(tmpdir(), 'hmrc-frontend-unit-tests-'));
   });
 
-  beforeAll(() => {
-    process.env.HMRC_FRONTEND_DISABLE_COMPATIBILITY_CHECK = '';
```
oscarduignan: basically you could just delete this beforeAll I think

oscarduignan: since the other stuff achieves the temp folder aspect sort of, though it's not using a temp folder just the current folder - but which is good enough as it's in a test folder and won't end up copied into build in any way

oscarduignan: just me unhelpfully backseat driving with that comment from PR 🙅

oscarduignan: but hopefully useful example in any case of tmpdir making stuff! though we only need to use tmpdirs since we're running subprocess, where most tests would mock the fs I think rather than creating actual files

TimothyFothergill (dismissed):
lgtm

PR #478, opened 2025-08-13
PLATUI-3901: adding in some more explicit waits to combat vrt test flakiness

oscarduignan:
haven't really been able to find a way to put my finger on the scale such that the flakiness we're experiencing becomes more reproducible so opting instead to just add in some explicit timeouts after actions we're experiencing flakiness with in a bit of a brute force attempt. I have omitted npm version bump to avoid a release with no functional changes.

### DIFF
--- a/backstop_data/engine_scripts/puppet/beforeTakingScreenshotHelper.js
+++ b/backstop_data/engine_scripts/puppet/beforeTakingScreenshotHelper.js
@@ -4,8 +4,11 @@ module.exports = async (page, { beforeTakingScreenshot }) => {
   allowedPageActions.click = async selector =>
     await page.click(selector);
 
-  allowedPageActions.hover = async selector =>
+  allowedPageActions.hover = async selector => {
     await page.hover(selector);
+    // to try and reduce seemingly unavoidable flakiness
+    await page.waitForTimeout(100);
+  }
 
   allowedPageActions.focus = async selector =>
     await page.focus(selector);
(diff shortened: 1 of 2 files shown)


backstop_data/engine_scripts/puppet/beforeTakingScreenshotHelper.js
```diff
@@ -16,8 +19,11 @@ module.exports = async (page, { beforeTakingScreenshot }) => {
   allowedPageActions.press = async key =>
     await page.press(key);
 
-  allowedPageActions.waitFor = async selector =>
-    await page.waitForSelector(selector);
+  allowedPageActions.waitFor = async selector => {
+    await page.locator(selector).waitFor();
```
oscarduignan, line 23: this is the recommended format for waiting, vs waitForSelector according to the docs of playwright

backstop_data/engine_scripts/puppet/beforeTakingScreenshotHelper.js
```diff
@@ -4,8 +4,11 @@ module.exports = async (page, { beforeTakingScreenshot }) => {
   allowedPageActions.click = async selector =>
     await page.click(selector);
 
-  allowedPageActions.hover = async selector =>
+  allowedPageActions.hover = async selector => {
     await page.hover(selector);
+    // to try and reduce seemingly unavoidable flakiness
+    await page.waitForTimeout(100);
```
oscarduignan, line 10: waitForTimeout is not recommended, but seems like it might be unavoidable for us - and we don't use these enough in our tests for it to add that much overhead at 100ms each use

src/components/account-header/account-header.yaml
oscarduignan, line 1: added in explicit waits for all tests that open the dropdown

oscarduignan: not enough by itself because we see issue happening after hover, and with the scenario we already implemented waitFor on - but it shows the intent clearly, and hopefully the 100ms waits are enough to make it no longer flaky in practice

oscarduignan:
I will try running this pull request in jenkins a number of times at see if I can trigger the flakiness with the changes 🤷‍♂️

PatrykRudnicki (approved):
LGTM

PR #480, opened 2025-08-15
PLATUI-3897 adds an upscaled png version of the govuk-crest

TimothyFothergill:
# Adds an upscaled png version of the govuk-crest

**New feature** 

- Adds govuk-crest-2x.png to the `src/assets/images` directory

## Checklist


oscarduignan:
could we add the instructions for how to generate this into the MAINTAINERS.md file alongside what you added for the previous PR:

```
## Double check if the govuk-crest needs updating

Under `src/assets/images/`, there is a png of the govuk-crest. For the most part, it's unlikely we will need to change this image, but at some point in the future, you may need to update this. This can be done by taking any software that can take and convert an svg file into the png format.

... link to the source svg

### How-to regenerate 1x and 2x versions from if SVG has been updated

... steps to create it 

### Why do we have these?

... explain why we are maintaining this so that future maintainers can check if it's still needed 
```

the "..." are bits to add, the first paragraph is what was there already

src/assets/images/govuk-crest-2x.png
oscarduignan, line 1: 😍

MAINTAINERS.md
```diff
@@ -87,7 +87,17 @@ Sometimes we may get deprecation warnings, but we might not always want to resol
...
+
+### How-to regenerate 1x and 2x versions from SVG if it has been updated
+
+If [govuk-frontend](https://github.com/alphagov/govuk-frontend/) provides an updated svg file for [govuk-crest](https://github.com/alphagov/govuk-frontend/blob/main/dist/assets/images/govuk-crest.svg), you can use any software to convert the file to png format. For vector-based illustrations (svg files), you could use [Inkscape](https://inkscape.org/), or any other vector-based software that allows exporting to the png file format.
+
+`govuk-crest-2x` may not be included in [govuk-frontend](https://github.com/alphagov/govuk-frontend/). If that's the case, you will need to take the govuk-crest file and upscale it yourself, using any vector-based software that allows you to do so. Then you will need to export it as a png file.
```
oscarduignan: this tells me what I need to do but not exactly how to do it 

since it's for us, we can be quite specific I think

I reckon if we assume maintainers will use inkscape then we can describe the steps for exporting it /  upscaling it then exporting it

for a how-to I think someone should be able to see it and follow the steps, like a numbered list of steps

oscarduignan: a specific example doesn't mean that people have to follow that, but makes it more likely if they translate steps to something else that they get it right

also makes it more likely that any unspoken conditions are easier to spot (if someone tries it in something else and results are wildly different) - like we have used inkscape, and maybe that does something specific like optimize image size or something like that which isn't called out explicitly but should probably also be done

oscarduignan (approved):

oscarduignan:
reaction-wow

oscarduignan:
merged just because I am going to rebase my branch for service nav 🙏

PR #485, opened 2025-09-10
PLATUI-3928 clean npm dependencies

TimothyFothergill:
# Clean npm dependencies

Initially: **65 vulnerabilities (3 low, 42 moderate, 15 high, 5 critical)**

### npm audit fix

Down to 54 vulnerabilities

### npm audit fix --force && npm dedupe

Drops us down to 10 vulnerabilities

Ran all tests and ran the application locally, all appears to be working as intended, so good first parse!

### babel-traverse

We were using an old plugin here:
```json
"test": {
	"plugins": ["babel-plugin-transform-es2015-modules-commonjs"]
}
```

But this has now been combined into one plugin which we pull as a dependency of something else anyway:
```json
"test": {
	"plugins": ["@babel/plugin-transform-modules-commonjs"]
}
```
Removed `babel-plugin-transform-es2015-modules-commonjs`

### form-data
Set an override for `^4.0.4` - That cleared this one up.

### tough-cookie
Set an override for `^4.1.3` - That cleared this one up.

### rollup
Bumped in devDependencies to `^4.50.1`, set override to `$rollup` to ensure dependencies use this.

Had to make some changes to allow us to move off deprecated plugins.

### request
Completely deprecated, removed and updated the one test file that utilised this to use fetch which is built in with js now

### adr-log
Nothing wrong with this one, but it had a number of dependencies and it required additional developer overhead to remember to use it, plus it added extra unnecessary newlines to the adr index file. Removed and this removed another 17 packages. Added some description in the MAINTAINERS.md file.

### html5shiv
We no longer support IE, this is now removed

### rimraf
Removed as unused

### mockdate
 Removed and using jest's built in setSystemTime with useFakeTimers.

### chalk
Removed and directly reference ansi colour codes for the same effect

## Result

From 1741 packages down to 1438 packages.
0 vulnerabilities.

## Small change to Welsh visually hidden text
- From 'English Newid yr iaith ir Gymraeg Cymraeg'
- To 'English Newid yr iaith i\'r Gymraeg Cymraeg'

## Checklist


package.json
```diff
@@ -74,35 +72,31 @@
...
     "puppeteer": "^23.4.1",
     "recursive-readdir": "^2.2.2",
     "remarkable": "^1.7.4",
-    "request": "^2.88.2",
     "rimraf": "^5.0.0",
-    "rollup": "^1.12.0",
-    "rollup-plugin-babel": "^4.4.0",
+    "rollup": "^4.50.1",
     "rollup-plugin-commonjs": "^10.1.0",
     "rollup-plugin-node-resolve": "^5.2.0",
```
oscarduignan: we've bumped rollup but not these - looks like node-resolve is on v16 - did we try updating and there was an issue? apologies if you already talked with me about it and my brain has forgotten

package.json
```diff
@@ -49,16 +48,15 @@
...
+    "@rollup/plugin-babel": "^6.0.4",
     "adm-zip": "^0.5.5",
-    "adr-log": "^2.2.0",
     "ajv": "^8.12.0",
-    "autoprefixer": "^9.6.1",
-    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
+    "autoprefixer": "^10.4.21",
     "backstopjs": "6.3.25",
     "better-npm-audit": "^3.7.3",
     "chalk": "^4.1.0",
```
oscarduignan: I can live without colourful messages when of log messages when starting vrt

Image

oscarduignan: (guess it might still come in transitively, but I think we can be pretty brutal with these - try to get rid of as much as possible, try to update as much stuff to recent major versions as possible)

package.json
```diff
@@ -74,35 +72,31 @@
...
     "jest": "^29.7.0",
     "jest-axe": "^3.5.0",
     "jest-environment-jsdom": "^29.7.0",
-    "jest-environment-node": "^29.7.0",
     "jest-puppeteer": "^11.0.0",
     "jest-serializer-html": "^5.0.0",
     "js-yaml": "^3.14.1",
     "mkdirp": "^3.0.1",
     "mockdate": "^3.0.2",
     "node-fetch": "^3.3.1",
```
oscarduignan, line 86: how are we using this one, is it this file https://github.com/hmrc/hmrc-frontend/blob/98a498ec5694a4eaf599edf0c438a02a9759ee04/backstop_data/engine_scripts/puppet/ignoreCSP.js#L4 seems like that file isn't being loaded by looking at backstop config, so we could remove that and the dep and see if backstop tests pass ok and if so 🙏

package.json
```diff
@@ -74,35 +72,31 @@
...
     "nunjucks": "^3.2.4",
     "playwright-chromium": "^1.49.1",
     "playwright-core": "^1.49.1",
     "postcss": "^8.0.0",
-    "postcss-pseudo-classes": "^0.2.0",
-    "postcss-scss": "^2.0.0",
+    "postcss-pseudo-classes": "^0.4.0",
+    "postcss-scss": "^4.0.9",
     "puppeteer": "^23.4.1",
     "recursive-readdir": "^2.2.2",
```
oscarduignan, line 95: I think we could get rid of this too

seems just used here https://github.com/hmrc/hmrc-frontend/blob/98a498ec5694a4eaf599edf0c438a02a9759ee04/tasks/gulp/__tests__/after-build-package.test.js#L20-L29

we could maybe refactor to use the glob package, which I think will do the same thing, and we also already pull in https://www.npmjs.com/package/glob

package.json
TimothyFothergill: There was a little bit of extra fiddling required to get those plugins to bump. Notice though that I did upgrade babel - That was as simple as swapping one out for the other (We now use `@rollup/plugin-babel`) - I don't mind trying to spend a bit more time to get those sorted too, prefer it tbh!

package.json
TimothyFothergill: I wasn't sure what to do with chalk here. I thought I'd leave it for this type of chat. If you think we're fine without chalk, I'm happy to drop it! 🫡

package.json
```diff
@@ -74,35 +72,31 @@
...
     "remarkable": "^1.7.4",
-    "request": "^2.88.2",
     "rimraf": "^5.0.0",
```
oscarduignan: I think we could remove rimraf, can't find it used anywhere in code from searching for rimraf - was going to say we could drop this or "del" as they do the same thing I think, but rimraf seems unused

package.json
TimothyFothergill: Okay so this was odd. When I delete or add jest-environment-node... No packages are added or removed? 🤷🏻

package.json
TimothyFothergill: Nice sounds good to me 👍🏻

package.json
TimothyFothergill: I was wondering which we wanted. I think, _think_ del required rimraf, but I might be wrong and will have a play around to figure if we can drop it!

package.json
```diff
@@ -74,35 +72,31 @@
...
     "husky": "^8.0.3",
     "jest": "^29.7.0",
     "jest-axe": "^3.5.0",
     "jest-environment-jsdom": "^29.7.0",
-    "jest-environment-node": "^29.7.0",
     "jest-puppeteer": "^11.0.0",
     "jest-serializer-html": "^5.0.0",
     "js-yaml": "^3.14.1",
     "mkdirp": "^3.0.1",
     "mockdate": "^3.0.2",
```
oscarduignan: jest has the ability to mock timers built in now so we can drop this and replace the single usage with https://jestjs.io/docs/jest-object#jestsetsystemtimenow-number--date

oscarduignan: even though it has no dependencies, it's still a liability

package.json
oscarduignan: time-say

package.json
oscarduignan: maybe there's something to split out then instead if they need fiddling to look at the outputs we're building with rollup and how - maybe compared to what govuk-frontend is doing - and generally refactor our build config as a whole

package.json
oscarduignan: maybe coming in transitively somehow

package.json
oscarduignan: tbh I wish we could get rid of both rimraf and del - and glob (but glob more widely used)

govuk-frontend is using node 22 now and that has

https://nodejs.org/api/fs.html#fsglobpattern-options-callback fs.glob

https://nodejs.org/api/fs.html#fsrmpath-options-callback this has been around for a while and docs say

> To get a behavior similar to the rm -rf Unix command, use [fs.rm()](https://nodejs.org/api/fs.html#fsrmpath-options-callback) with options { recursive: true, force: true }.

oscarduignan: since a lot of this stuff is used in tests / build scripts - we can be pretty brutal, but if something looks too big we can spin up another ticket and do it in sequence

src/components/header/header.yaml
```diff
@@ -673,7 +673,7 @@ examples:
                 </li>
                 <li class="hmrc-service-navigation-language-select__list-item">
                   <a href="/components/service-navigation-language-select/welsh/preview" hreflang="cy" lang="cy" rel="alternate" class="govuk-link govuk-service-navigation__link" data-journey-click="link - click:lang-select:Cymraeg">
-                    CYM<span class="govuk-visually-hidden"> – Newid yr iaith ir Gymraeg</span>
+                    CYM<span class="govuk-visually-hidden"> – Newid yr iaith i'r Gymraeg</span>
```
oscarduignan: should it be a normal ' or slightly angled one ’ ? Some other places we have the slightly angled one https://github.com/hmrc/hmrc-frontend/blob/98a498ec5694a4eaf599edf0c438a02a9759ee04/src/components/accessible-autocomplete/accessible-autocomplete.js#L49

```suggestion
                    CYM<span class="govuk-visually-hidden"> – Newid yr iaith i’r Gymraeg</span>
```

TimothyFothergill: Just pushed a change, they now use `’` instead of `'` 😵

TimothyFothergill:
Ran into issues when rebasing, closing this PR and raising a new one

PR #494, opened 2025-10-14
PLATUI-3998 bump govuk-frontend and remove sass-mq refs

TimothyFothergill:
# Bump govuk-frontend version and remove sass-mq refs

**Version bump** 

- Bump govuk-frontend to latest (5.13.0)
- Remove all references to sass-mq

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-## [6.96.0] - 2025-10-14
+## [6.97.0] - 2025-10-14
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [6.96.0] - 2025-10-14
+
+### Changed
+
+- Bump govuk-frontend version
+- Remove all usage of sass-mq
+
 ## [6.96.0] - 2025-10-06
 
 ### Changed
--- a/src/components/account-menu/_account-menu.scss
+++ b/src/components/account-menu/_account-menu.scss
@@ -358,7 +358,7 @@ $hmrc-banner-blue: #005ea5;
   }
 }
 
-@media (min-width: govuk-media-query(tablet)) {
+@include govuk-media-query($from: tablet) {
   .hmrc-full-width-banner__close {
     right: 35px;
   }
--- a/src/components/account-menu/_account-menu.scss
+++ b/src/components/account-menu/_account-menu.scss
@@ -358,7 +358,7 @@ $hmrc-banner-blue: #005ea5;
   }
 }
 
-@media (min-width: govuk-breakpoint-value(tablet)) {
+@media (min-width: govuk-media-query(tablet)) {
   .hmrc-full-width-banner__close {
     right: 35px;
   }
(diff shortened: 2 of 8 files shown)


src/components/account-menu/_account-menu.scss
```diff
@@ -359,7 +358,7 @@ $hmrc-banner-blue: #005ea5;
   }
 }
 
-@include mq(tablet) {
+@media (min-width: govuk-breakpoint-value(tablet)) {
```
oscarduignan: I think we should avoid using this because it was only added 2 weeks ago, so using this will mean that new versions of hmrc-frontend are incompatible with all but the latest govuk-frontend https://github.com/alphagov/govuk-frontend/commit/c769d66643afa787fba973efb25ab6c622887bbf

I think we could still use govuk-media-query to do it, which is what they are deprecating I think (and replaced under the hood maybe with using govuk-breakpoint-value) https://github.com/alphagov/govuk-frontend/commit/c769d66643afa787fba973efb25ab6c622887bbf#diff-653f88427bb6e3d2c1f473121bf98706af24c887510b3aa2ecc80f679bf11475L52-L76

oscarduignan: I'm not sure why we were using sass-mq directly, might be a legacy thing where at one point it was supposed to be used directly then they added an internal interface for it so they could change it but we didn't migrate

oscarduignan: I peaked at govuk-frontend and it looked like it was there at least 7 years, but then peaked at a hmrc-frontend use too and we were using it for at least 7 years too so probably one of those - it worked fine for 7 years so 🤷‍♂️ cant' really argue with it xD

src/components/account-menu/_account-menu.scss
```diff
@@ -359,7 +358,7 @@ $hmrc-banner-blue: #005ea5;
   }
 }
 
-@include mq(tablet) {
+@media (min-width: govuk-media-query(tablet)) {
```
oscarduignan: I'm surprised to see this work because in govuk-frontend it seems to suggest usage like:

```
@include govuk-media-query($from: tablet) {
  color: green;
}
```

if you package this locally does it work ok in a prototype using an older version of govuk-frontend? what does the generated css look like (I guess fine since VRT tests passed?)

oscarduignan: think I need to read the change from govuk more closely - maybe I have missed something

oscarduignan (dismissed):

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [6.96.0] - 2025-10-14
```
oscarduignan: sorry, only just spotted this!

```suggestion
## [6.97.0] - 2025-10-14
```

src/govuk-prototype-kit.config.json
```diff
@@ -10,7 +10,7 @@
   "pluginDependencies": [
     {
       "packageName": "govuk-frontend",
-      "minVersion": "5.12.0"
+      "minVersion": "5.13.0"
```
oscarduignan, line 13: I don't understand what this minVersion does still, and if we should be bumping it when we bump the govuk-frontend version or leaving it at where we are compatible back to

oscarduignan: not said to prompt any action to be taken, just stating that I'm still confused xD

oscarduignan (approved):

PatrykRudnicki (approved):

PR #496, opened 2025-10-23
PLATUI-4002 Add CSS to support long service name

PatrykRudnicki:
# Descriptive Title

Fixes # (issue)

## Checklist

* [X] I've read the [CONTRIBUTING](../CONTRIBUTING.md) guidance, including next steps and [expected response](../CONTRIBUTING.md#when-can-i-expect-someone-to-look-at-my-external-contribution) from the owners
* [X] I've added appropriate unit tests, and run all [unit tests](../CONTRIBUTING.md#unit-tests)
* [X] I've run the [visual regression test using Backstop](../CONTRIBUTING.md#visual-regression-tests) and updated the images if needed
* [X] I've run `npm version minor` to update the version in [package.json](../package.json) and [package-lock.json](../package-lock.json)
* [X] I've updated the [CHANGELOG](../CHANGELOG.md)


src/components/header/_header.scss
```diff
@@ -169,3 +169,13 @@ $govuk-header-link-underline-thickness: 3px;
 .govuk-template--rebranded .hmrc-header + .hmrc-user-research-banner {
   top: 0;
 }
+
+.hmrc-header--with-support-for-long-service-names .govuk-header__content {
+  display: flex;
```
oscarduignan, line 176: I think we need `align-items: end` here too so that the sign out link has the same baseline as the last line of the header, so a bit like this:

Image

PatrykRudnicki: Hmm, I thought that the expected behaviour is a sign out link aligned to the top, first title line, but sure, happy to add it too

oscarduignan: and then I think it was this alignment that I was varying the line height to get it to align the same across viewports - might be a better way to achieve that with some playing around though that varying line height across breakpoints

PatrykRudnicki: Okay, thanks. Will play with this

oscarduignan: ah, I think in one of the slack thread I shared different examples of it - where one it was aligned at the top - and I think there was some what-if from me around how service header from one login works - where I think in theirs the sign out link is centered vertically rather than top of bottom, and answer for now was:

> IMO, with the current sign out pattern the baseline of 'Sign out' should be aligned with the baseline of the lowest line in the header. But longer term we'll most likely align to whatever One Login and now doing.

oscarduignan:
I need to resolve conflicts but also spotted I have a dud VRT test captured, a next week problem to fixup, but otherwise I think it's in a good state - fixes some other inconsistencies that we'd kind of patched around so css should be bit simpler now - have made it so that it doesn't need the extra class added to fix stuff in favour of using :has() and :not() combination of things which I think should be fairly sound - if we aren't happy with that I'd be in favour of something like an extra class to disable the good default behaviour rather than not supporting common use case of component I think - I might be way off here though and in full friday brain obliviousness

image

oscarduignan:
have rebased this in new pr because I made some changes that benefited from squashing and I didn't want to accidentally blow away stuff - this pr is replaced by #501

PR #501, opened 2025-11-10
Plat UI 4002 make sure that sign out link and service name stay on same baseline

oscarduignan:
reworked and squashed #496 to make rebase easier

# hmrc-header: make sure that sign out link and service name stay on same baseline even if the service name is long enough to go onto multiple lines

**Bug fix**

We had considered a new class for this but instead we can make it work with some modern CSS - which is ok because we're covering edge cases here that won't impact most people - and in the future people will be using service navigation for links and their service name rather than the header so this won't be an issue past a certain point

## Checklist

* [x] Need to add Patryk as co-author (will do on squash / merge)


backstop_data/bitmaps_reference/backstop_default_add-to-a-list_mutltiple-generic-welsh-items_example_0_document_0_phone.png
oscarduignan, line 1: as part of this I removed some examples that we were no longer generating

separate to this I'll create a maintainer doc with how I did this

(where I generated reference images and then compared list of files in the folders and deleted the differences because backstop doesn't come with this out of the box)

package.json
oscarduignan, line 1: I kept running into some issues with backstop in this where it was chrome via playwright was hanging - either because of the app getting blocked or the browser in container I'm not sure ( I think maybe it was the container because docker seemed to lock up ) I tried running server in separate process and increasing resources but nothing seemed too reliable - have some ideas for how to improve our backstop config in future which might help I think

src/components/header/_header.scss
oscarduignan, line 1: combination of padding and line height to position sign out link was making it very difficult to position things before - instead just matching the margin bottom of the service name and using flex to `align-items: last end;` I will test how this looks in internet explorer but we still have the float left / right in here so flex is just to get bottom alignment - would probably like to hand this over to someone like @timsb to check I haven't forgotten some edge case here

src/components/header/_header.scss
```diff
@@ -169,3 +156,46 @@ $govuk-header-link-underline-thickness: 3px;
...
+.hmrc-header:has(.hmrc-sign-out-nav):not(:has(.govuk-header__navigation)) .hmrc-sign-out-nav {
+  flex-shrink: 0;
+  margin-left: auto;
+}
+
+.hmrc-header:has(.hmrc-sign-out-nav):not(:has(.govuk-header__navigation)) .hmrc-sign-out-nav__link {
+  margin-left: 10px;
+}
+
+.hmrc-header:has(.govuk-header__navigation) .hmrc-sign-out-nav {
```
oscarduignan, line 187: probably like a handful at most of services if even that

src/components/header/_header.scss
```diff
@@ -169,3 +156,46 @@ $govuk-header-link-underline-thickness: 3px;
...
+}
+
+.hmrc-header:has(.hmrc-sign-out-nav):not(:has(.govuk-header__navigation)) .hmrc-sign-out-nav {
+  flex-shrink: 0;
+  margin-left: auto;
+}
+
+.hmrc-header:has(.hmrc-sign-out-nav):not(:has(.govuk-header__navigation)) .hmrc-sign-out-nav__link {
+  margin-left: 10px;
+}
```
oscarduignan: not sure why this can't just go on `.hmrc-sign-out-nav__link` all the time - the extra selectors I might have just copied out of habit and aren't actually needed

backstop_data/bitmaps_reference/backstop_default_header_sign_out_link_in_welsh_example_0_document_1_tablet.png
oscarduignan, line 1: the difference on these is that aligning by last baseline changes position by 1px still when there's not multiple line service - figured that isn't a problem is likely is that we were 1px off before

src/components/header/_header.scss
```diff
@@ -12,30 +12,18 @@ $govuk-header-link-underline-thickness: 3px;
 
   .hmrc-sign-out-nav {
     float: right;
+    margin-bottom: 10px;
 
     .govuk-template--rebranded & {
-      @include govuk-media-query($from: desktop) {
-        padding-top: govuk-spacing(3) + 3px; // makes the baseline match gov.uk logo and service name
```
oscarduignan, line 18: rather than being handled by top padding, now this will work via `display: flex` + `align-items: last baseline`

src/components/header/_header.scss
```diff
@@ -12,30 +12,18 @@ $govuk-header-link-underline-thickness: 3px;
...
-      }
-    }
-  }
-
-  .govuk-header__navigation + .hmrc-sign-out-nav {
-    .govuk-template--rebranded & {
-      padding-top: 0;
-      padding-bottom: govuk-spacing(2);
```
oscarduignan, line 26: same as above, no more positioning by padding

oscarduignan: there's a few tweaks to positioning of this through margin top at the bottom of file now

feel like I want to double check how it looks with rebrand / no rebrand after this though before tomorrow

src/components/header/_header.scss
oscarduignan: I've refactored this file a bit to trim stuff from my initial proposition, because I was away and then came back and thought to myself "you have created trash oscar" bit tidier now

CHANGELOG.md
```diff
@@ -44,7 +50,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - Update npm exceptions list
+<<<<<<< HEAD
 - 
+=======
+
+>>>>>>> 3a4c764 (PLATUI-4002: stop long service names moving sign out link to new line)
```
timsb: There's some Git conflict markers here. Guessing you want to remove them?

src/components/header/_header.scss
timsb: Yeah this makes sense to position from bottom as you're now aligning to bottom

timsb (dismissed):
This all looks better to me and the consistent spacing — using margin-bottom — to control the position makes it easier to understand what's happening

CHANGELOG.md
oscarduignan: 🤦‍♂️ thank you tim!

PatrykRudnicki (approved):

PR #516, opened 2026-02-11
NOJIRA fix compatibility issue with govuk-frontend v6

oscarduignan:
We spotted this a while ago (because of broken compat with older govuk frontend versions) and then forgot about it!

:facepalm: totally my fault

# stop using mixin not available in old / latest govuk-frontend

**Bug fix**

with this mixin being used we aren't compatible with older versions of govuk-frontend or now the latest release

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [6.114.0] - 2026-02-11
+
+### Changed
+
+- Removed use of _govuk-rebrand mixin to be compatible with govuk-frontend v6
+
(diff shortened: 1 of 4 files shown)


oscarduignan:
@nataliecarey thank you for raising this - you explained this exactly and I thought "we expected not to be compatible" but actually we didn't expect not to be compatible in this way and I didn't look closely enough at what you raised 😮‍💨 

I was just testing our pinning of prototype kit version for new prototypes and ran into it too and realised it was an easily fixable thing, that we'd even spotted before and planned to fix (because it also broke backwards compatibility) but had gotten lost

will @ attribute you in slack message PSAing this 🙏  and let me know if you want me to add an attribution any other way for raising it

PatrykRudnicki (approved):

PR #517, opened 2026-02-13
PLATUI-4161: Update to govuk-frontend v6.0.0

JoPintoPaul:
# Descriptive Title

Fixes # (issue)

## Checklist


tasks/gulp/backstop-config.js
```diff
@@ -71,10 +71,7 @@ module.exports = ({ host, port, components }) => ({
   ],
   onBeforeScript: 'playwright/onBefore.js',
   onReadyScript: 'playwright/onReady.js',
-  scenarios: buildScenarioList(host, port, components).flatMap((scenario) => [
-    // this means we have twice as many VRT which might increase test time a lot...
-    scenario, { ...scenario, label: `${scenario.label} (using rebrand)`, url: `${scenario.url.replace('preview.html', 'preview-rebrand.html')}` },
-  ]),
```
oscarduignan, line 77: I think we should do this in 2 changes, otherwise VRT tests won't be useful here where since we're going to be styling stuff in the header it would be good to have the VRT tests for the rebranded header to work

we could drop the bit that varies the url from this, then we will still capture a non-rebranded and a rebranded example

the non-rebranded ones will now be rebranded

and the rebranded ones will capture the changes in the header

then in next release we can stop generating the (using rebrand) ones

src/components/service-navigation-language-select/_service-navigation-language-select.scss
```diff
@@ -52,7 +52,7 @@
     margin-bottom: 0;
     padding: govuk-spacing(4) 0;
 
-    .govuk-template--rebranded & {
+    .govuk-template & {
       padding: govuk-spacing(3) 0;
```
oscarduignan, line 56: I think for these the change would be to fold this in as the default, so it would replace what's on line 53

src/components/header/template.njk
```diff
@@ -3,112 +3,55 @@
...
-      <a href="{{ params.homepageUrl | default("/", true) }}" class="govuk-header__link govuk-header__link--homepage">
-        {{ govukLogo({
-          classes: "govuk-header__logotype",
-          ariaLabelText: "GOV.UK",
-          useTudorCrown: params.useTudorCrown,
-          rebrand: _rebrand
-        }) | trim | indent(8) }}
-        {% if (params.productName) %}
-        <span class="govuk-header__product-name">{{ params.productName }}</span>
+<header class="govuk-template__header">
```
oscarduignan, line 11: I guess this is the question, do we keep <header> in hmrcHeader or remove it (so govukHeader and hmrcHeader are interchangeable still)

govuk team have added extra param to layout so what you used to pass to header (output of `govukHeader({ ... })`) you should now pass to govukHeader block instead

I am not sure if we would add that extra param in play-frontend-hmrc too or just tell people to make sure they add <header> if they are overriding the default header block when using templates - feels a bit confusing to have both things, and it would probably still be easy for people to do nothing and end up with accessibility issues when <header> disappears if they don't change to pass it via new param - vs just saying "you must start wrapping govukHeader in <header> if you use it directly"

🤷‍♂️

PR #525, opened 2026-03-04
PLATUI-4192: Updated to use govuk-frontend v6.1.0

JoPintoPaul:
# Uplift to use govuk-frontend v6.1.0

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [7.1.0] - 2026-03-04
+
+### Changed
+
+- Uplifted to use `govuk-frontend` v6.1.0
+- Updated division in `.scss` files to use `sass:math`, for example replaced `(29 / 16)` with `math.div(29, 16)`
+
 ## [7.0.0] - 2026-03-02
 
 ### Changed
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,17 +1,17 @@
 {
   "name": "hmrc-frontend",
-  "version": "7.0.0",
+  "version": "7.1.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "hmrc-frontend",
-      "version": "7.0.0",
+      "version": "7.1.0",
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
         "accessible-autocomplete": "^3.0.1",
-        "govuk-frontend": "^6.0.0"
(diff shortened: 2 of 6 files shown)


src/govuk-prototype-kit.config.json
```diff
@@ -10,7 +10,7 @@
   "pluginDependencies": [
     {
       "packageName": "govuk-frontend",
-      "minVersion": "6.0.0"
+      "minVersion": "6.1.0"
```
oscarduignan: I think this might be able to stay at 6.0.0

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [7.1.0] - 2026-03-04
+
+### Changed
+
+- Uplifted to use `govuk-frontend` v6.1.0
+- Updated division in `.scss` files to use `calc`, for example replaced `(29 / 16)` with `calc(29 / 16)`
```
oscarduignan: on this calc change, why did we need to do this - I couldn't spot where it came from

JoPintoPaul: @oscarduignan this was done due to an error we've been seeing for a while:
```
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div(29, 16) or calc(29 / 16)

More info and automated migrator: https://sass-lang.com/d/slash-div
```

Seeing as we've now dropped `node-sass`, this seemed like a good time to finally make this change (we couldn't before as neither of those commands worked with `node-sass`, IIRC).

TimothyFothergill (dismissed):
lgtm

CHANGELOG.md
oscarduignan: I wonder if we should use math.div() instead, that would be more close to what we were doing before I think - math.div() would calculate it in sass and output a number, like (29 / 16), whereas calc() will be output to css and browser will interpret the calculation

my yard stick is always WWGDSD (what would govuk design system do) I can see govuk-frontend use it for some https://github.com/alphagov/govuk-frontend/blob/42e77788ed7cf3e53ab0c5c2bc75d5b36d724a39/packages/govuk-frontend/src/govuk/components/service-navigation/_index.scss#L57

they seem to save calc for stuff that has dynamic value like a css var https://github.com/search?q=repo%3Aalphagov%2Fgovuk-frontend+%2Fcalc%5C%28%2F&type=code

JoPintoPaul: @oscarduignan updated to use `sass:math`

timsb (approved):

oscarduignan (approved):

PR #529, opened 2026-03-23
Caseworker banner

timsb:
# Caseworker guidance banner

**New feature**

This adds the Caseworker guidance banner which provides concise, essential instructions for caseworkers in internal HMRC services.

It is based on the govuk-frontend notificaion banner with some parameters set and styling overidden to look distinct from other notification banners.

Due to the fact that the text parameter has a bold font-weight when used, the caseworker guidance text param has had to be passed to the notification banner text param with an added `<p class="govuk-body"></p>` element wrapping it.

## Checklist


backstop_data/bitmaps_reference/backstop_default_caseworker-banner_text_content_example_using_rebrand_0_document_1_tablet.png
Jon-Rowe-HMRC, line 1: @timsb the text wrapping is strange on this example

timsb: Yeah this is because it sits outside of `.govuk-width-container`. The notification banner has some CSS that sets a max-width of 605px across all content. I presume this is so that reading line lengths aren't too long. This shouldn't happen with our banners as the guidance says to add the banner under the h1 which will be constrained by the `.govuk-width-container`. You can see the same things in the govuk notification banner examples e.g. https://design-system.service.gov.uk/components/notification-banner/default/

Jon-Rowe-HMRC (approved):

src/components/caseworker-banner/template.njk
```diff
@@ -0,0 +1,26 @@
...
+  disableAutoFocus: true,
+  html: content,
+  role: null,
+  text: null,
+  titleHeadingLevel: 2,
+  titleHtml: null,
+  titleId : params.titleId,
+  titleText: titleText,
+  type: null
+}) }}
```
oscarduignan, line 27: looks all good to me, but to say out loud to check- the nulls are to be clear pattern is intended not to pass something for these, so someone recreating know not to allow passing these through

not sure if it's worth it but we could have a comment like `{# explicit nulls here are to show that caseworker banner is only intended to use a subset of notification banner options #}`

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [7.3.0] - 2026-03-23
+
+### Added
+
+Caseworker banner component
+
```
oscarduignan: I guess this will be a good test of if anyone gets caught out by being stuck on hmrc-frontend v6 due to needing to migrate to service nav and not being able to yet - and wanting to use new stuff - where I can imagine people might be like "oh well, can't get those new accessibility fixes" sometimes but if it's like "wait you mean we can't have the new component?!?" that might be different story - at least it will be easy enough I guess if someone does need it for them to copy over the styles and build html themselves (where backporting it to hmrc-frontend v6 is probably more work than needed for something that might never be used)

oscarduignan (dismissed):

src/components/caseworker-banner/template.njk
timsb: Yep, that's right. It probably seems more necessary because we don't have the parameters displayed in the examples like GOV.UK Design System does so the code is the only place to look.

oscarduignan (approved):

PR #541, opened 2026-05-14
PLAUI-4290: Added min-release-age of 7 days and ignore-scripts to npm…

JoPintoPaul:
# Add `ignore-scripts` and `min-release-age` of 7 days to `npm` configuration

**New feature**

Adding in the following for additional security against supply chain attacks in `npm`:
* `min-release-age` of 7 days
* `ignore-scripts` set to `true`
* `allow-git` set to `none`

Additionally, as scripts can no longer execute from `npm` packages, prior to running `jest src` we now manually install the Playwright and Puppeteer browsers via build-jobs and the new `scripts` command.

## Checklist


PatrykRudnicki (dismissed):
LGTM

package.json
```diff
@@ -16,13 +16,14 @@
     "build": "npm run build:dist && npm run build:package",
     "lint": "stylelint 'src/**/*.scss' && eslint .",
     "lint:fix": "stylelint 'src/**/*.scss' --fix && eslint --fix .",
-    "test": "npm run build:dist && npm run test:compatibility && jest src && npm run lint",
+    "test": "npm run build:dist && npm run test:compatibility && npm run install:playwright && npm run lint",
```
oscarduignan: ah, we'll still need to run `jest src` here - jest is like scalatest, so this is like sayings "run all tests in src folder"

I think we should not automatically run browser install on test, but have it as separate task, since we mostly just need it for remote build environment

package.json
```diff
@@ -16,13 +16,14 @@
...
     "test:update-snapshots": "jest src -u",
     "test:backstop": "gulp backstopTest",
     "test:backstop-approve": "gulp backstop:approve",
     "audit": "better-npm-audit audit",
     "prepare": "husky install",
-    "postinstall": "node check-compatibility-with-govuk-frontend-version.js"
+    "postinstall": "node check-compatibility-with-govuk-frontend-version.js",
+    "install:playwright": "playwright install"
```
oscarduignan: interesting, I'm surprised we don't need to install stuff for puppeteer too - maybe puppeteer runs it's installation not as a lifecycle script but inline as part of the using it

package.json
```diff
@@ -16,13 +16,14 @@
...
     "lint:fix": "stylelint 'src/**/*.scss' --fix && eslint --fix .",
-    "test": "npm run build:dist && npm run test:compatibility && jest src && npm run lint",
+    "test": "npm run build:dist && npm run test:compatibility && npm run install:playwright && npm run lint",
     "test:update-snapshots": "jest src -u",
     "test:backstop": "gulp backstopTest",
     "test:backstop-approve": "gulp backstop:approve",
     "audit": "better-npm-audit audit",
     "prepare": "husky install",
-    "postinstall": "node check-compatibility-with-govuk-frontend-version.js"
+    "postinstall": "node check-compatibility-with-govuk-frontend-version.js",
```
oscarduignan: I think as part of adding ignore-scripts, we can remove this

looks like prototype kit now disables scripts, where I think this was supposed to be run - so not useful anymore since with scripts disabled it won't run

https://github.com/alphagov/govuk-prototype-kit/blame/e76b78b96804d48480ae18a6ec418ebc1607c49e/bin/cli#L66-L70

package.json
```diff
@@ -22,7 +22,8 @@
     "test:backstop-approve": "gulp backstop:approve",
     "audit": "better-npm-audit audit",
     "prepare": "husky install",
-    "postinstall": "node check-compatibility-with-govuk-frontend-version.js"
```
oscarduignan, line 25: I think following this, we can remove check-compatibility-with-govuk-frontend-version.js script altogether I guess - I don't think we suggest anyone to run it manually - I think it was a good intention but now we're in this post-lifecycle-script world

TimothyFothergill (dismissed):
lgtm

package.json
JoPintoPaul: @oscarduignan we're still running it manually as part of the `test` script:
```
"test": "npm run build:dist && npm run test:compatibility && jest src && npm run lint",
```

Are you thinking to drop that?

oscarduignan: ah I think `npm run test:compatibility` is running the tests we have to make sure that `node check-compatibility-with-govuk-frontend-version.js` would work, but if we remove the postinstall, then we can remove the script and the tests which were verifying it worked

https://github.com/hmrc/hmrc-frontend/blob/33076d52c53d401c2ee6d4a88a3104eded44d328/package.json#L7

it's also in the build https://github.com/hmrc/hmrc-frontend/blob/33076d52c53d401c2ee6d4a88a3104eded44d328/gulpfile.js#L30 and section of readme that explained how to disable it https://github.com/hmrc/hmrc-frontend/blob/33076d52c53d401c2ee6d4a88a3104eded44d328/README.md?plain=1#L34-L40

oscarduignan: I think maybe the gulpfile.js needs to copy task invocation removed too and the place it's implemented too in tasks/gulp/package.js

oscarduignan (dismissed):

PatrykRudnicki (dismissed):

oscarduignan (dismissed):

PatrykRudnicki (dismissed):

TimothyFothergill (approved):
lgtm

PR #555, opened 2026-07-20
PLATUI-4385: Uplifted govuk-frontend to v6.4.0

JoPintoPaul:
# Uplifted govuk-frontend to v6.4.0

**BAU**

## Checklist


src/components/caseworker-banner/_caseworker-banner.scss
```diff
@@ -1,4 +1,7 @@
 .hmrc-caseworker-banner {
   border-color: govuk-colour("black");
-  background-color: govuk-colour("black");
+
+  .govuk-notification-banner__header {
+    background-color: govuk-colour("black");
```
oscarduignan, line 5: I think it's a future thing to pick up with @timsb because I'm not sure, but this also made me think - we aren't explicitly setting the color of text, we're relying on inherited value being appropriate for the background color

wonder if we should set it explicitly, but from my pov this makes sense - gds moved background color onto __header and so we have done the same

oscarduignan (approved):

PR #567, opened 2026-09-11
PLATUI-4401: Adding align inline to language toggle in service navigation

JoPintoPaul:
# Adding align inline to language toggle in service navigation

**Bug fix**

Adding align inline to language toggle in service navigation.

Fixes # (issue)

## Checklist


src/components/service-navigation-language-select/layout.njk
```diff
@@ -23,7 +23,10 @@
         }
       ],
       slots: {
-        end: componentView
+        end: {
+          html: componentView,
+          align: "inline"
```
oscarduignan: I guess we probably want an example of it not aligned inline so that we can check it still displays correctly

where I inline [adds extra styles to container in service nav](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/service-navigation/template.njk#L17)  which we do already with asking people to add class of their own

but I guess switching to inline means we can drop that need - but should probably still be compatible if that's passed / people haven't passed inline

oscarduignan: and maybe one where the hmrc class is passed to service nav still

JoPintoPaul: @oscarduignan so looking at what happens when we pass in `hmrc-service-navigation--with-language-select` to the `ServiceNavigation`?
https://github.com/hmrc/play-frontend-hmrc/blob/main/play-frontend-hmrc-play-30/src/main/scala/uk/gov/hmrc/govukfrontend/views/implicits/RichServiceNavigationSupport.scala#L40

JoPintoPaul: @oscarduignan Just checked, the example alyout does currently have that class in it:
https://github.com/hmrc/hmrc-frontend/blob/PLATUI-4401_language-toggle-alignment-fix/src/components/service-navigation-language-select/layout.njk#L9 - so we know what happens with the class there - I can see what happens when we remove it too

backstop_data/bitmaps_reference/backstop_default_service-navigation-language-select_welsh_example_focused_0_document_0_phone.png
oscarduignan, line 1: This seems different to the one above, but they are both the same width - only difference seems to be focus styles - I'm guessing it shouldn't move when focused

Image

oscarduignan: the font size on the added one below seems bit bigger too to me

backstop_data/bitmaps_reference/backstop_default_service-navigation-language-select_with_long_name_with_long_words_example_hovered_0_document_0_phone.png
oscarduignan, line 1: so for this one, the service name would be out of viewport and need to be scrolled to be seen in full but not the language select - just saying out loud my understand of this screenshot

backstop_data/bitmaps_reference/backstop_default_header_with_service_navigation_and_govukHeader_example_0_document_0_phone.png
oscarduignan, line 1: @timsb @JoPintoPaul I've noticed in some of these that the colour of the separator in between the ENG | CYM and text size I think of ENG | CYM hasn't updated in all the examples - think it might be that our visual difference is set too low to pickup the change in VRT screenshots for them

oscarduignan: we might be able to set it to be more sensitive for this - accept the changes - the change sensitivity back to be less sensitive to avoid rendering differences between local machines / jenkins causing test failures there in build (not ideal I know)

backstop_data/bitmaps_reference/backstop_default_service-navigation-language-select_welsh_example_focused_0_document_0_phone.png
Jon-Rowe-HMRC, line 1: On the newer screen the padding above the toggle seems too much, and too little below.

My thoughts:
The padding below the baseline of ENG should match the padding above the top-line of Example.
The padding above the top-line of ENG and baseline of Menu should match the padding between the top-line of Menu and baseline of Example.

Jon-Rowe-HMRC (commented):
I don't mind if the pipe is grey (like the GDS language navigation) or black (like the HMRC welsh toggle). But if we go with the former we should change https://design.tax.service.gov.uk/hmrc-design-patterns/welsh-language-toggle/ too.

backstop_data/bitmaps_reference/backstop_default_service-navigation-language-select_welsh_example_focused_0_document_0_phone.png
timsb: @JoPintoPaul I can't see this locally. Is it an old image that's not been removed for some reason?

JoPintoPaul: @timsb I can see this image on my machine. It looks like still a valid test case, `welsh` and `focused` - you can see this is in the `main` branch at the moment:
https://github.com/hmrc/hmrc-frontend/blob/main/src/components/service-navigation-language-select/service-navigation-language-select.yaml#L42

backstop_data/bitmaps_reference/backstop_default_header_with_service_navigation_and_govukHeader_example_0_document_0_phone.png
JoPintoPaul: @timsb and @oscarduignan as there was some odd behaviour going on around VRT images, I've now pushed up two separate commits. In the first, I checked out existing reference images from `main` and deleted newly added ones. In a second commit, I then updated the reference images with created or modified ones. Could you both take a look at the reference images at some point?

backstop_data/bitmaps_reference/backstop_default_service-navigation-language-select_welsh_example_focused_0_document_0_phone.png
JoPintoPaul: @oscarduignan and @timsb - got to the bottom of the VRT issue, it was due to me making a mistake in the YAML VRT syntax - all images should be corrected now.

backstop_data/bitmaps_reference/backstop_default_header_with_service_navigation_and_govukHeader_example_0_document_0_phone.png
JoPintoPaul: @oscarduignan fixed now, see above.

Jon-Rowe-HMRC (approved):

PatrykRudnicki (approved):

